PR #3, opened 2026-06-09
[contra] Reserve upper_bound headroom for active re-state in has_deposit_slot

### LINKED ISSUE #2: update_active_balance can overflow EncryptedBalance.upper_bound and lock pending deposits
## Summary

`update_active_balance` re-states an account's `active` balance, which resets `active.upper_bound` to 1 (via `overwrite`) without re-checking the combined active+pending growth budget that `has_deposit_slot` reserves. On a freshly registered account (`active.upper_bound == 0`) whose `pending` is full and whose public balance is non-zero, the next `merge` overflows `upper_bound` (u16) and aborts permanently, which locks the pending and public deposits.

## Details

`EncryptedBalance.upper_bound` (u16) counts the values folded into a balance and bounds limb growth so the balance stays within the decryption window. `merge` folds pending and the public balance into active:

```move
acc.active.merge_into(&mut acc.pending);            // active.upper_bound += pending.upper_bound
acc.active.merge_public(acc.public_balance.take()); // active.upper_bound += 1 (when public > 0)
```

`has_deposit_slot` gates deposits so that `active.upper_bound + pending.upper_bound <= max_upper_bound() - 1` (`0xFFFE`), reserving exactly one slot for the `merge_public` `+1`. A `merge` from a freshly registered account (`active.upper_bound == 0`) therefore tops out at `max_upper_bound()` (`0xFFFF`), which is safe.

`update_active_balance` is the only path that raises `active.upper_bound` outside `merge`/spend without re-checking that budget: on a fresh account it goes from 0 to 1. With `pending.upper_bound == max_upper_bound() - 1` (`0xFFFE`) and a non-zero public balance, the next `merge` computes `1 + 0xFFFE + 1 = 0x10000`, which overflows the u16 `upper_bound` and aborts. Every subsequent `merge` aborts identically, so the pending and public deposits can never be folded into active and cannot be spent (spend draws only from active). The funds are permanently locked. The account can also no longer receive deposits (the slot is full) or rotate its key (re-keying requires an empty pending).

## Reproduction (all valid operations)

1. Register an account (`active.upper_bound == 0`).
2. Receive encrypted transfers until `pending.upper_bound == max_upper_bound() - 1` (`0xFFFE`).
3. Receive one `wrap` so the public balance is non-zero.
4. Call `update_active_balance` (`active.upper_bound` goes from 0 to 1).
5. Every `merge` now aborts with an arithmetic error.

The documented flow (merge first, then `update_active_balance`) is unaffected, because after a merge `pending` is empty.

## Suggested fix

Require `pending.is_empty()` in `update_active_balance` before re-stating, matching `set_public_key_internal` (which re-states `active` under the same precondition) and the documented use of the function ("update the balance after merging many pending deposits").

wstran:
## Summary

`merge` folds pending and the public balance into active, raising `active.upper_bound` by `pending.upper_bound` plus 1 (the `merge_public` bump). `has_deposit_slot` reserved one `upper_bound` slot for that bump, so a `merge` from a freshly registered account topped out at `max_upper_bound()`.

But re-stating active (`update_active_balance` or `set_public_key`) resets `active.upper_bound` to 1 outside the deposit path. With a full `pending`, the next `merge` computed `1 + (max_upper_bound() - 1) + 1`, overflowing `upper_bound` (u16) and aborting permanently, which locks the pending and public deposits.

This reserves a second slot in `has_deposit_slot` (`pending.upper_bound() < max_upper_bound() - 2`) so the worst-case `merge` after a re-state (`1 + pending + 1`) stays within `max_upper_bound()`.

Fixes #2.

## Notes

- Per review, the reservation now lives in `has_deposit_slot` rather than guarding `update_active_balance`, so re-stating is not over-restricted (it no longer requires an empty `pending`).
- The cap is tight: with `pending.upper_bound()` held at `max_upper_bound() - 2`, the worst-case merge lands exactly on `max_upper_bound()`.

## Tests

- `update_active_balance_allows_unmerged_pending`: re-stating with an unmerged pending deposit succeeds (no longer over-restricted).
- `update_active_balance_succeeds_with_empty_pending`: the documented re-state flow still works.
- `merge_at_deposit_cap_does_not_overflow_upper_bound`: at the deposit cap, a re-state plus a full pending plus the public bump lands exactly on `max_upper_bound()`.
- `merge_one_past_deposit_cap_overflows_upper_bound`: one past the cap (now unreachable) the merge overflows, showing the reservation is necessary.

`sui move test`: 27 passed.

### DIFF
--- a/move/sources/contra.move
+++ b/move/sources/contra.move
@@ -971,13 +971,16 @@ fun has_token<T>(account: &Account): bool {
     df::exists(&account.id, TokenAccountKey<T>())
 }
 
-/// Slots available for new pending deposits. Always reserves one slot for a possible future
-/// `merge_public` bump, so the cap compared against is `max_upper_bound() - 1` rather than
-/// `max_upper_bound()`.
+/// Slots available for new pending deposits. Reserves two `upper_bound` headroom slots beyond
+/// `pending`: one for a future `merge_public` `+1`, and one for the `+1` that re-stating `active`
+/// (`update_active_balance` or `set_public_key`) adds when it resets `active.upper_bound` to 1
+/// outside the deposit path. Holding `pending.upper_bound()` at `max_upper_bound() - 2` at most
+/// keeps the worst-case `merge` after such a re-state (`1 + pending.upper_bound() + 1`) within
+/// `max_upper_bound()`, so the u16 `upper_bound` never overflows.
 fun has_deposit_slot<T>(self: &TokenAccount<T>): bool {
     let cap = balance::max_upper_bound_minus_1();
     let used = self.active.upper_bound() + self.pending.upper_bound();
-    cap > used
+    cap > used && self.pending.upper_bound() < balance::max_upper_bound() - 2
 }
 
 /// 20-byte session_id for `account`'s `TokenAccount<T>`.
--- a/move/sources/balance.move
+++ b/move/sources/balance.move
@@ -312,3 +312,20 @@ fun set_zero<T>(self: &mut EncryptedBalance<T>) {
     self.amount = encrypted_amount::zero();
     self.upper_bound = 0;
 }
+
+// === Test helpers ===
+
+#[test_only]
+public(package) fun new_with_upper_bound_for_testing<T>(upper_bound: u16): EncryptedBalance<T> {
+    EncryptedBalance { amount: encrypted_amount::zero(), upper_bound }
+}
+
+#[test_only]
+public(package) fun public_coin_for_testing<T>(value: u64): PublicCoin<T> {
+    PublicCoin { value }
+}
+
+#[test_only]
+public(package) fun destroy_for_testing<T>(self: EncryptedBalance<T>) {
+    let EncryptedBalance { amount: _, upper_bound: _ } = self;
+}
--- a/move/tests/contra_tests.move
+++ b/move/tests/contra_tests.move
@@ -6,6 +6,7 @@ module contra::contra_tests;
 
 use contra::{
     auditors,
+    balance,
     contra,
     encrypted_amount::{Self, consistency_proof_for_testing},
     nizk,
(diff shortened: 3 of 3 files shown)


move/sources/contra.move
```diff
@@ -706,6 +712,7 @@ public fun update_active_balance<T>(
     assert!(auth.is_authenticated(account.owner), EAuthorizationError);
     let owner = account.owner;
     let token_account = &mut account[TokenAccountKey<T>()];
+    assert!(token_account.pending.is_empty(), EPendingDepositsMustBeMerged);
```
jonas-lj: I think this is too strict - we just need `pending.upper_bound() < max_upper_bound_minus_1()` here or, alternatively, change `has_deposit_slot` function to check that `cap > used + 1`.

jonas-lj (commented):
Thanks for contributing!

move/sources/contra.move
jonas-lj: I prefer the latter, or better yet, to add a check in `has_deposit_slot` that `pending.upper_bound() < max_upper_bound() - 2`

wstran:
Thanks @jonas-lj. Agreed, `pending.is_empty()` was stricter than necessary. I went with your preferred option: the reservation now lives in `has_deposit_slot` (`pending.upper_bound() < max_upper_bound() - 2`), and I reverted the `update_active_balance` guard. That keeps the worst-case merge after a re-state bumps `active.upper_bound` to 1 (`1 + pending + 1`) at exactly `max_upper_bound()`.

Updated the tests: re-stating with unmerged pending now succeeds, plus white-box tests that the merge `upper_bound` lands on `max_upper_bound()` at the deposit cap and would overflow one past it. Pushed.

jonas-lj:
After discussing this with the team, we ended up choosing a different solution, https://github.com/MystenLabs/confidential-transfers/pull/4, so I'm closing this PR, but thanks so much for pointing this out - good catch!

wstran:
Thanks @jonas-lj, appreciate the fast turnaround. For me what matters is the best fix landing, not whose PR it is, so I'm genuinely glad PR #4 is the one going in. Starting `active` at `upper_bound = 1` removes the `ub = 0` edge case at the source, which is cleaner than the slot reservation I proposed. Good to see it sorted before mainnet, and happy to help if anything else comes up.

PR #4, opened 2026-06-09
Initialize active with upper_bound = 1; rename balance::new → balance::empty

jonas-lj:
Initialize `active` with `upper_bound = 1` (via a new `balance::new`. The old `new` is renamed to `balance::empty` and used for `pending`). With `active.ub ≥ 1` as a permanent invariant, `has_deposit_slot` caps `pending.ub ≤ 0xFFFD`, so a later `update_active_balance` / `set_public_key_internal` restate (which lands at `active.ub = 1`) can't push the next `merge` past the u16 budget.

Alternative to the per-call assert in #3: closes the same bug upstream, no runtime check, no false-positive race against incoming transfers between read and execution.


benr-ml (dismissed):
pls remove all the comments that start from "Used by... "

benr-ml (approved):

PR #9, opened 2026-06-11
Clean up decryption/dlog implementation

jonas-lj:
Clean up the dlog implementation. Half the size of the intermediate result from `computeTableEntries` .

Using arrays instead of a `Map` reduces the memory footprint of the table by x4.5 for numBits = 16 (6.7 -> 1.5 MB) and x11 for larger numBits. This is due to the overhead in the representation of `Map`.

### DIFF
--- a/ts-sdk/src/twisted_elgamal.ts
+++ b/ts-sdk/src/twisted_elgamal.ts
@@ -166,7 +166,7 @@ export class DiscreteLogTable {
 	 */
 	static fromEntries(numBits: number, entries: Uint32Array): DiscreteLogTable {
 		const n = entries.length;
-		
+
 		const values = new Uint32Array(n);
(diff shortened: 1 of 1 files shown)


ts-sdk/src/twisted_elgamal.ts
```diff
@@ -59,6 +59,11 @@ function cosetX(point: RistrettoPoint): bigint[] {
 	return shifted.map((s: any, k) => Fp.mul(s.X, invertedZs[k]));
 }
 
+/** Key used for indexing the discrete-log table. Computed from the  */
```
benr-ml:
```suggestion
/** Key used for indexing the discrete-log table. */
```

benr-ml (dismissed):

benr-ml (dismissed):

benr-ml (approved):

PR #10, opened 2026-06-21
[SDK] more robust interfaces

benr-ml:


ts-sdk/src/auditor.ts
```diff
@@ -87,7 +94,17 @@ export class ContraAuditor {
 			}
 			return mrc.decrypt(entry.index, entry.privateKey, this.#table);
 		});
-		return limbsToScalar(limbs);
+		// The on-chain key-consistency proof only binds the reconstructed limbs to the account
+		// public key modulo the scalar order `q`, so reduce to the canonical scalar.
+		const recoveredKey = limbsToScalar(limbs) % SCALAR_ORDER;
```
jonas-lj: If this actually reduces it means the user constructed non-canonical limbs. This shouldn't happen for an honest user, so perhaps it should be logged here?

ts-sdk/src/ristretto255.ts
```diff
@@ -59,6 +59,9 @@ H.precompute(8);
 /** The ristretto255 identity (zero) point. */
 export const ZERO: RistrettoPoint = ristretto255.Point.ZERO;
 
+/** The order `q` of the ristretto255 scalar field (number of scalars). */
```
jonas-lj:
```suggestion
/** The order of the ristretto255 scalar field. */
```

ts-sdk/src/ristretto255.ts
```diff
@@ -59,6 +59,9 @@ H.precompute(8);
...
+export const SCALAR_ORDER: bigint = ristretto255.Point.Fn.ORDER;
```
jonas-lj: Maybe call this GROUP_ORDER instead?

ts-sdk/src/auditor.ts
```diff
@@ -87,7 +94,17 @@ export class ContraAuditor {
 			}
 			return mrc.decrypt(entry.index, entry.privateKey, this.#table);
 		});
-		return limbsToScalar(limbs);
+		// The on-chain key-consistency proof only binds the reconstructed limbs to the account
+		// public key modulo the scalar order `q`, so reduce to the canonical scalar.
```
jonas-lj:
```suggestion
		// public key modulo the group order, so reduce to the canonical representation.
```

jonas-lj (approved):

PR #13, opened 2026-06-24
Re-key balances with a batched DDH proof

jonas-lj:
Replaces the key-rotation flow's new-balance well-formedness proof with a batched re-keying DDH (rekey): the client sends only the 4 new decryption handles and proves a single witness re-keys the public key and every limb's decryption handle from the old key to the new one. Commitments are reused from the existing balance, so no range/consistency proofs are needed.


move/sources/nizk.move
```diff
@@ -67,6 +73,13 @@ public fun new_ddh_proof(a: Element<G>, b: Element<G>, z: Element<Scalar>): DdhP
     DdhProof { a, b, z }
 }
 
+public fun new_batched_ddh_proof(
+    commitments: vector<Element<G>>,
+    z: Element<Scalar>,
+): BatchedDdhProof {
+    BatchedDdhProof { commitments, z }
```
benr-ml, line 80: should we verify we have exactly 5 commitments?

ts-sdk/src/client.ts
```diff
@@ -863,25 +864,23 @@ export class ContraClient {
...
-			const proof = ElGamalNizk.prove(elgamalDst, l.blinding, l.value, ciphertext, newPk);
-			return { value: l.value, blinding: l.blinding, ciphertext, proof };
-		});
+		// Per-limb re-key under the NEW public key, reusing each limb's `(value, r)`. The Pedersen
+		// commitment is unchanged; only the decryption handle changes from `r * oldPk` to `r * newPk`.
+		// No range/consistency proof is needed — the batched DDH below proves each handle is the
+		// matching old handle re-keyed by the shared witness, and `try_set_public_key` checks the
+		// commitments are byte-identical.
+		const newBalanceUnderNewPk = newBalanceUnderOldPk.map((l) => ({
+			ciphertext: new Ciphertext(l.ciphertext.ciphertext, mul(newPk, l.blinding)),
```
benr-ml: how do we have r here? we can only assume we know sk, sk'

ts-sdk/src/client.ts
```diff
@@ -1364,16 +1360,16 @@ export class ContraClient {
...
-			return { value: l.value, blinding: l.blinding, ciphertext, proof };
-		});
-		const rCollapsed = collapseBlindings(restateUnderOldPk);
-		const oldHandle = mul(oldPk, rCollapsed);
-		const newHandle = mul(newPk, rCollapsed);
+		// Per-limb re-key under the NEW public key: reuse each limb's commitment, only the decryption
+		// handle changes from `r*oldPk` to `r*newPk`. The batched DDH below proves the re-keying, so no
+		// range/consistency proof is needed.
+		const balanceUnderNewPk = restateUnderOldPk.map((l) => ({
+			ciphertext: new Ciphertext(l.ciphertext.ciphertext, mul(newPk, l.blinding)),
```
benr-ml: also here

move/sources/nizk.move
jonas-lj: The count is already enforced in `verify_batched_ddh` since it requires `commitments.length()` to equal the number of base/image pairs

ts-sdk/src/client.ts
jonas-lj: Switched to deriving the new handles as w * D_i where w = newSk/oldSk applied to the old on-chain handles)

ts-sdk/src/client.ts
jonas-lj: Same fix here

ts-sdk/src/client.ts
```diff
@@ -863,25 +864,19 @@ export class ContraClient {
...
-		});
-
-		// DDH proof for `try_set_public_key`: knowledge of `w = newSk * oldSk^{-1}` such that
-		// `w * oldPk = newPk` AND `w * oldHandle = newHandle`. Combined with the byte-equality check
-		// on ciphertexts that try_set_public_key enforces, this attests the new balance has the same
-		// collapsed blinding `r` as the old. Witness is derived from the user's two secret keys.
-		const rCollapsed = collapseBlindings(newBalanceUnderOldPk);
-		const oldHandle = mul(oldPk, rCollapsed);
-		const newHandle = mul(newPk, rCollapsed);
+		// Batched re-key (Π_rekey) for `try_set_public_key`: the single witness `w = newSk * oldSk^{-1}`
```
benr-ml: can shorten the doc

ts-sdk/src/client.ts
benr-ml, line 847: let's add another proto id, e.g. PROTOCOL_BATCH_DDH

benr-ml (approved):

PR #14, opened 2026-06-24
Optimize sender amount encoding in batched transfers

jonas-lj:
Batched transfers no longer send per-receiver sender-keyed amounts — only the receiver amounts plus one sender decryption handle for the transfer total.
The sender recovers its own outgoing amounts from the commitments by re-deriving per-limb blindings from `seed = HKDF(sk * P)`, where `P` (`seed_point`) is published in each `TransferEvent`.


ts-sdk/src/twisted_elgamal.ts
```diff
@@ -267,15 +294,17 @@ export class Ciphertext {
...
+	 * `blinding` defaults to a fresh random scalar; pass it explicitly to re-key
+	 * an existing amount while keeping the same per-limb commitment, or to use a
+	 * deterministically derived blinding (e.g. a batched transfer's
+	 * seed-derived per-limb blindings).
 	 */
 	static encryptWithConsistencyProof(
 		dst: Uint8Array,
 		pk: PublicKey,
 		value: bigint,
+		blinding: bigint = randomScalar(),
```
benr-ml: let's remove the default and mke it explicit

benr-ml (approved):

PR #24, opened 2026-07-31
[Move] Per-transaction auditing

jonas-lj:
Switches auditing from per-account viewing-key escrow to per-transfer.

Also adds a default_pk to `Account` (shared across tokens) which can be used to deposit tokens to an account without requiring the user to register for that token first (assuming registration is permissionless).


move/sources/contra.move
```diff
@@ -152,12 +168,15 @@ public struct Pool<phantom T> has key {
 public struct Account has key {
     id: UID,
     owner: address,
+    pk: Element<G>,
+    accepts_deposits: bool,
```
benr-ml: do we need it in the account level?

jonas-lj: No — `accepts_deposits` is now a per-`TokenAccount` field (the deposit gate is per token, and after a rotation a token can keep receiving under its old key while stale). `Account` only holds the shared account key `pk`, which is the account-level convergence target for rotation.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

move/sources/contra.move
```diff
@@ -152,12 +159,12 @@ public struct Pool<phantom T> has key {
 public struct Account has key {
     id: UID,
     owner: address,
+    default_key: Option<Element<G>>,
```
benr-ml:
```suggestion
    default_pk: Option<Element<G>>,
```

move/sources/contra.move
```diff
@@ -293,16 +302,42 @@ public fun share_confidential_token<T>(ct: ConfidentialToken<T>) {
...
-///
-/// Note: the `owner` argument is not tied to `ctx.sender()` — anyone can create an
-/// `Account` on behalf of any address. Since `Account` has `key` only (no `store`),
-/// the only way to dispose of it outside this module is via `share_account`, and
-/// all authenticated operations still gate on `account.owner == ctx.sender()`.
-public fun new_account(registry: &mut AccountRegistry, owner: address): Account {
+/// Create a new account for `ctx.sender()` with optional `default_key` (see `Account`). Creation
+/// is restricted to the owner: only the sender can set their own default key. Pass `none` to create
+/// without one (no third party can then auto-register tokens for this account). Can only happen once
+/// per address. For an account owned by a Move object, use `new_account_for_object`.
```
benr-ml: remove 
>Can only happen once
/// per address. For an account owned by a Move object, use `new_account_for_object`.

move/sources/contra.move
```diff
@@ -293,16 +302,42 @@ public fun share_confidential_token<T>(ct: ConfidentialToken<T>) {
...
+public fun new_account(
+    registry: &mut AccountRegistry,
+    default_key: Option<Element<G>>,
```
benr-ml:
```suggestion
    default_pk: Option<Element<G>>,
```

move/sources/contra.move
```diff
@@ -293,16 +302,42 @@ public fun share_confidential_token<T>(ct: ConfidentialToken<T>) {
...
+    ctx: &TxContext,
+): Account {
+    new_account_internal(registry, ctx.sender(), default_key)
+}
+
+/// Create a new account owned by the object identified by `uid`, with optional `default_key`.
+/// Holding `&mut UID` proves custody of the object, so it self-authenticates as its own owner (the
+/// address derived from the `UID`) — mirroring `authorize_as_object`. Can only happen once per object
```
benr-ml: remove

> — mirroring `authorize_as_object`. Can only happen once per object
/// address.

move/sources/contra.move
```diff
@@ -293,16 +302,42 @@ public fun share_confidential_token<T>(ct: ConfidentialToken<T>) {
...
+/// address.
+#[allow(unused_mut_parameter)]
+public fun new_account_for_object(
+    registry: &mut AccountRegistry,
+    uid: &mut UID,
+    default_key: Option<Element<G>>,
```
benr-ml:
```suggestion
    default_pk: Option<Element<G>>,
```

move/sources/contra.move
```diff
@@ -312,55 +347,45 @@ public fun share_account(account: Account) {
...
-}
-
-public(package) fun register_internal<T>(
-    account: &mut Account,
-    auth: &Auth<T>,
-    pk: Element<G>,
-    verified_key_encryption: VerifiedKeyEncryption,
-    session_id: vector<u8>,
-) {
+/// Create a `TokenAccount` for token `T`, with its balances keyed under `pk`. Since per-token keys
```
benr-ml: remove from 
>Since per-token keys

move/sources/contra.move
```diff
@@ -312,55 +347,45 @@ public fun share_account(account: Account) {
...
+    account.add_token_account<T>(pk, session_id);
+}
 
-    events::emit_new_registration<T>(account.owner, pk, verified_key_encryption);
+/// Permissionless counterpart to `register`: create a `TokenAccount<T>` on `account` on behalf of its
+/// owner, keyed under the account's `default_key`, without any `Auth`. Only allowed when
+/// `T` leaves registration permissionless and `default_key` is set. Aborts
+/// if `default_key` is unset (`EDefaultKeyNotSet`), `T`'s registration is permissioned
+/// (`ERegistrationNotPermissionless`), or the token is already registered (`EAccountAlreadyRegistered`).
+public fun register_permissionless<T>(account: &mut Account, ct: &ConfidentialToken<T>) {
```
benr-ml:
```suggestion
public fun register_with_default_pk<T>(account: &mut Account, ct: &ConfidentialToken<T>) {
```

move/sources/contra.move
benr-ml: let's remove too verbose docs everywhere

move/sources/contra.move
```diff
@@ -385,127 +410,94 @@ public fun set_accepts_encrypted_deposits<T>(
...
+/// account is registered for.
+public fun set_default_key<T>(
     account: &mut Account,
     auth: &Auth<T>,
-    ct: &ConfidentialToken<T>,
-    new_pk: Element<G>,
-    new_handles: vector<Element<G>>,
-    rekey_proof: DdhProof,
-    key_encryption: Option<KeyEncryption>,
+    new_default_key: Option<Element<G>>,
```
benr-ml:
```suggestion
    default_pk: Option<Element<G>>,
```

move/sources/contra.move
```diff
@@ -385,127 +410,94 @@ public fun set_accepts_encrypted_deposits<T>(
...
-///   prevent new encrypted deposits.
-/// - Call `set_public_key` to update the public key and `set_accepts_encrypted_deposits` to true
-///   to allow new encrypted deposits again.
-public fun set_public_key<T>(
+/// Set the account's optional `default_key` (pass `none` to clear it, which disables permissionless
+/// auto-registration for this account). This is purely the key `register_permissionless` uses;
+/// per-token keys are unaffected and are rotated independently via `rekey_token`. Authenticated by
+/// `auth`, which must be for the `PERMISSIONED_REGISTER` operation and may be for any token the
+/// account is registered for.
+public fun set_default_key<T>(
```
benr-ml: why is this function generic?
also, it should not use Auth but the account/obj as create account functions do

move/sources/events.move
```diff
@@ -51,6 +70,8 @@ public struct TransferEvent<phantom T> has copy, drop {
     receiver: address,
     receiver_pk: Element<G>,
     encrypted_amount_receiver: EncryptedAmount,
+    auditor_handles: vector<Element<G>>,
```
benr-ml: let's have a type of EncryptedAmountAuditorHandles/AuditorHandles?

move/sources/nizk.move
```diff
@@ -93,9 +58,14 @@ public(package) fun verify_ddh(
     bases: &vector<Element<G>>,
     images: &vector<Element<G>>,
 ): bool {
-    // TODO: check for degenerate case where a base is the identity element.
     let n = bases.length();
     if (images.length() != n || proof.commitments.length() != n) return false;
+    // Reject a fully degenerate statement — every base the identity — which binds no witness. An
```
benr-ml: is this comment correct? why do we need it?

move/sources/auditors.move
```diff
@@ -3,216 +3,107 @@
...
+    if (auditor.current_pk.is_some_and!(|pk| proof.verify_elgamal(dst, pk, encryptions))) {
+        return auditor.current_pk
     };
+    if (
+        epoch <= auditor.previous_expiration_epoch &&
+            auditor.previous_pk.is_some_and!(|pk| proof.verify_elgamal(dst, pk, encryptions))
+    ) {
+        return auditor.previous_pk
+    };
+    option::none()
```
benr-ml: is there a case we shoud not fail for none? maybe use assert and return Element<G>

apps/closed-loop/move/sources/confidential_pbu.move
```diff
@@ -97,23 +97,24 @@ public fun is_whitelisted(whitelist: &Whitelist, addr: address): bool {
...
     account: &mut Account,
-    pk: Element<G>,
     ctx: &mut TxContext,
 ) {
     assert!(whitelist.addresses.contains(&ctx.sender()), ENotWhitelisted);
     // Auth is only used internally.
     let auth = ct.authorize_with_witness(REGISTER_OP, account.owner(), PbuWitness {});
-    contra::register(account, &auth, ct, pk, option::none());
+    // Key the token account under the account's default key (set at account creation).
+    let pk = *contra::default_key(account).borrow();
```
benr-ml: no, it should still receive pk, and ignore the default_pk

apps/closed-loop/move/sources/confidential_pbu.move
```diff
@@ -128,13 +129,6 @@ public fun set_public_key(
...
-    contra::set_public_key(
-        account,
-        &auth,
-        ct,
-        new_pk,
-        new_handles,
-        rekey_proof,
-        option::none(),
-    );
+    contra::set_default_key(account, &auth, option::some(new_pk));
```
benr-ml: why does it change account?

move/sources/nizk.move
jonas-lj: The comment was correct but overclaimed, so I tightened it. The guard rejects an all-identity statement and is defensive only. Every real statement includes a non-identity base (`g`, or the asserted-non-identity old key on re-key), and an individual identity base must stay allowed (e.g. for a new balance's identity limb).

move/sources/auditors.move
jonas-lj: Done — `verify_transfer` now asserts and returns `Element<G>` (aborts `EAuditorProofFailed`). The only caller already required success when data was attached, so there was no `none` case worth preserving.

apps/closed-loop/move/sources/confidential_pbu.move
jonas-lj: Removed. `set_public_key` now only re-keys the token and leaves the account default key untouched.

apps/payment-channel/move/payment_channel/sources/payment_channel.move
```diff
@@ -142,6 +142,19 @@ public fun new<T>(ctx: &mut TxContext) {
     transfer::share_object(c);
 }
 
+/// Create the contra `Account` owned by the channel's address, with account key `pk`. Restricted to
+/// the channel `sender`, and object-bound (the channel self-authenticates via its own `&mut UID`), so
+/// no one else can squat the channel's account with a key they control.
+public fun new_account<T>(
+    c: &mut Channel<T>,
+    registry: &mut AccountRegistry,
+    pk: Element<G>,
```
benr-ml: pk is not needed for this application, let's create without a default key

move/sources/auditors.move
```diff
@@ -4,215 +4,118 @@
...
+const EMissingAuditorData: u64 = 2;
+const EUnexpectedAuditorData: u64 = 3;
 
-// === Main Type(s) ===
+// === Main Type ===
 
-/// Holds the set of auditor `public_keys` registered for a token. Auditors can decrypt the
-/// viewing-key ciphertexts attached to each transfer, giving them read access to transaction
-/// amounts without being able to move funds.
+/// The auditor configuration for a confidential token under the per-transfer auditing model: a
```
benr-ml:
```suggestion
/// The auditor configuration for a confidential token: a
```

move/sources/auditors.move
```diff
@@ -4,215 +4,118 @@
...
-    version: u32,
-    recommended_min_version: u32,
-}
-
-/// A user's viewing key encrypted to each auditor's public key, stored on their account after
-/// passing a `KeyConsistencyProof` check. The `version` records which auditor key set it was
-/// produced against, so callers can compare it against `Auditors.recommended_min_version` to
-/// detect encryptions that the issuer considers stale.
+/// Auditing is per-transfer — every transfer carries auditor-readable ciphertexts of the amount
+/// (see `verify_transfer`) — so the auditor never learns the user's viewing key and balances are
```
benr-ml:
```suggestion
```

move/sources/auditors.move
```diff
@@ -4,215 +4,118 @@
...
+/// encrypted only under the user's own key.
```
benr-ml:
```suggestion
```

move/sources/auditors.move
```diff
@@ -4,215 +4,118 @@
...
 ///
-/// An empty `ciphertext` means the user's account was registered while the token had no
-/// auditors set.
-public struct VerifiedKeyEncryption has copy, drop, store {
-    ciphertext: vector<MultiRecipientEncryption>,
-    version: u32,
+/// `current_pk == none` means auditing is disabled: transfers must carry no auditor data. On
```
benr-ml:
```suggestion
/// `current_pk == none` means auditing is disabled: transfers may carry no auditor data. On
```

move/sources/auditors.move
```diff
@@ -4,215 +4,118 @@
...
-/// - `proof` is the `KeyConsistencyProof` showing each limb of `ciphertext` correctly
-///   encrypts the matching 32-bit limb of the user's private key under every auditor's
-///   public key, and that the limbs sum to `sender_public_key`'s discrete log.
-/// - `range_proof` is an aggregate Bulletproof showing every limb's plaintext lies
-///   in `[0, 2^32)` so that auditors can recover each limb via baby-step giant-step.
-public struct KeyEncryption has drop {
-    ciphertext: vector<MultiRecipientEncryption>,
-    proof: KeyConsistencyProof,
-    range_proof: vector<u8>,
+/// The per-transfer auditor data a sender attaches to a `batched_transfer`: one `U32LimbHandles`
```
benr-ml:
```suggestion
/// The per-transfer auditor data a sender attaches to a `batched_transfer`
```
als remove the next lines (they just describe the struct)

move/sources/contra.move
```diff
@@ -17,15 +17,16 @@
...
 ///    `sui::coin::deny_list_v2_add` / `sui::coin::deny_list_v2_remove`. The deny list affects
 ///    both the public and the private coin; to freeze only the private coin, see items 2 and 3.
-/// 6. Rotate or disable the auditor key set via `update_auditors` (using the ManagementCap).
-///    Setting `bump_recommended_min` raises the auditors' `recommended_min_version` to the new
-///    version, signalling to wallets that every user should refresh their key.
-///    Passing an empty `public_keys` vector disables the auditor flow.
+/// 6. Rotate or disable the auditor key via `update_auditor` (using the ManagementCap). The
+///    outgoing key stays valid for transfers through a caller-set `expiration_epoch` (a grace
+///    window for in-flight transfers). Passing `none` disables auditing going forward. Auditing is
+///    per-transfer and each transfer carries auditor-readable ciphertexts of the amount, so the auditor
```
benr-ml:
```suggestion
///    per-transfer and each transfer carries auditor-readable ciphertexts of the amount.
```

move/sources/contra.move
```diff
@@ -17,15 +17,16 @@
...
+///    never learns users' viewing keys.
```
benr-ml:
```suggestion
```

move/sources/contra.move
```diff
@@ -185,6 +195,8 @@ public enum TransferBatch<phantom T> {
         coins: vector<EncryptedCoin<T>>,
         seed_point: Element<G>,
         next_index: u8,
+        auditor_handles: vector<U32LimbHandles>,
```
benr-ml:
```suggestion
        auditor_decryption_handles: vector<U32LimbHandles>,
```

benr-ml: U32LimbHandles is very generic - if we use it only for the auditor, should we call it `AuditorDecryptionHandles`?

move/sources/contra.move
```diff
@@ -293,16 +305,38 @@ public fun share_confidential_token<T>(ct: ConfidentialToken<T>) {
...
-/// Create a new account for the given address. Can only happen once per address.
-///
-/// Note: the `owner` argument is not tied to `ctx.sender()` — anyone can create an
-/// `Account` on behalf of any address. Since `Account` has `key` only (no `store`),
-/// the only way to dispose of it outside this module is via `share_account`, and
-/// all authenticated operations still gate on `account.owner == ctx.sender()`.
-public fun new_account(registry: &mut AccountRegistry, owner: address): Account {
+/// Create a new account for `ctx.sender()` with optional `default_pk` (see `Account`). Pass `none`
+/// to create without one (no third party can then auto-register tokens for this account).
+public fun new_account(
```
benr-ml: new_account_as_sender (similar to authorize_as_sender)

move/sources/contra.move
```diff
@@ -293,16 +305,38 @@ public fun share_confidential_token<T>(ct: ConfidentialToken<T>) {
...
+    default_pk: Option<Element<G>>,
+    ctx: &TxContext,
+): Account {
+    new_account_internal(registry, ctx.sender(), default_pk)
+}
+
+/// Create a new account owned by the object identified by `uid`, with optional `default_pk`. Holding
+/// `&mut UID` proves custody of the object, so it self-authenticates as its own owner.
+#[allow(unused_mut_parameter)]
+public fun new_account_for_object(
```
benr-ml: new_account_as_object

move/sources/contra.move
benr-ml: what about new_account() that sets it without the deafult pk and can be called by anyone?

move/sources/contra.move
```diff
@@ -293,16 +305,38 @@ public fun share_confidential_token<T>(ct: ConfidentialToken<T>) {
...
+fun new_account_internal(
+    registry: &mut AccountRegistry,
+    owner: address,
+    default_pk: Option<Element<G>>,
+): Account {
     assert!(!derived_object::exists(&registry.id, AccountKey(owner)), EAccountAlreadyRegistered);
+    assert!(default_pk.is_none() || *default_pk.borrow() != g_identity(), EIdentityPublicKey);
     let id = derived_object::claim(&mut registry.id, AccountKey(owner));
-    Account { id, owner }
+    Account { id, owner, default_pk }
```
benr-ml: check that default_pk != identity

move/sources/contra.move
```diff
@@ -385,127 +403,99 @@ public fun set_accepts_encrypted_deposits<T>(
...
-/// authorization since the same flow gates account onboarding.
-/// This aborts if there are pending deposits that need to be merged, so the caller should:
-/// - Call `merge` to merge pending deposits and `set_accepts_encrypted_deposits` to false to
-///   prevent new encrypted deposits.
-/// - Call `set_public_key` to update the public key and `set_accepts_encrypted_deposits` to true
-///   to allow new encrypted deposits again.
-public fun set_public_key<T>(
+/// Set the account's optional `default_pk` (pass `none` to clear it, which disables permissionless
+/// auto-registration).
+public fun set_default_pk(account: &mut Account, default_pk: Option<Element<G>>, ctx: &TxContext) {
```
benr-ml:
```suggestion
public fun set_default_pk_as_sender(account: &mut Account, default_pk: Option<Element<G>>, ctx: &TxContext) {
```

move/sources/contra.move
```diff
@@ -385,127 +403,99 @@ public fun set_accepts_encrypted_deposits<T>(
...
+    assert!(ctx.sender() == account.owner, EAuthorizationError);
+    account.set_default_pk_internal(default_pk);
+}
+
+/// Set the `default_pk` of an account owned by the object identified by `uid`. Holding `&mut UID`
+/// proves custody of the object, so it self-authenticates as its own owner.
+#[allow(unused_mut_parameter)]
+public fun set_default_pk_for_object(
```
benr-ml:
```suggestion
public fun set_default_pk_as_object(
```

move/sources/contra.move
```diff
@@ -385,127 +403,99 @@ public fun set_accepts_encrypted_deposits<T>(
...
-                sid.dst(DST_RANGE_PROOF_32),
-            ),
-        sid.dst(DST_BATCH_DDH),
-    );
+    assert!(uid.to_inner().to_address() == account.owner, EAuthorizationError);
+    account.set_default_pk_internal(default_pk);
+}
+
+fun set_default_pk_internal(account: &mut Account, default_pk: Option<Element<G>>) {
+    assert!(default_pk.is_none() || *default_pk.borrow() != g_identity(), EIdentityPublicKey);
```
benr-ml: we have a function for that in auditors

move/sources/contra.move
```diff
@@ -385,127 +403,99 @@ public fun set_accepts_encrypted_deposits<T>(
...
-/// `TrySetPublicKeyFailedEvent` and leaves the account paused for a retry. The caller must `merge`
-/// (and pause) first.
-public fun try_set_public_key_and_unpause<T>(
+/// Re-key token `T`'s active balance from its current `TokenAccount.pk` to `new_pk`, swapping each
+/// limb's decryption handle for the matching `new_handles[i]` (proven by `rekey_proof`). `new_pk` is
+/// explicit and independent of the account's default key. Aborts if `new_pk` is the group identity,
+/// the token has unmerged pending deposits (which are under the old key, so they must be merged
+/// first), or the proof fails. Authorized by `auth`, which must be for the `PERMISSIONED_REGISTER`
+/// operation and for `account.owner`.
+public fun rekey_token<T>(
```
benr-ml:
```suggestion
public fun rekey_token_account<T>(
```

move/sources/contra.move
```diff
@@ -385,127 +403,99 @@ public fun set_accepts_encrypted_deposits<T>(
...
+public fun try_rekey_token<T>(
     account: &mut Account,
     auth: &Auth<T>,
     new_pk: Element<G>,
     new_handles: vector<Element<G>>,
     rekey_proof: DdhProof,
-    new_verified_key_encryption: VerifiedKeyEncryption,
-    dst: vector<u8>,
-) {
+): bool {
```
benr-ml: do we need to return bool?

move/sources/contra.move
```diff
@@ -385,127 +403,99 @@ public fun set_accepts_encrypted_deposits<T>(
...
+    }
+}
+
+/// Shared re-key: Assert `new_pk` is non-identity and the token's pending is empty, then re-key its
+/// active balance from `TokenAccount.pk` to `new_pk`. On a verifying proof, commits the new handles,
+/// sets `token.pk = new_pk`, emits `TokenRekeyedEvent`, and returns `true`. Otherwise leaves the token
+/// unchanged and returns `false`.
+fun rekey_token_internal<T>(
+    account: &mut Account,
+    new_pk: Element<G>,
```
benr-ml: maybe let's define a type NonTrivialElement(Element<G>) with 
public non_trivial_element(g: Element<G>): NonTrivialElement { check not identity}

and use it everywhere instead of those 
assert!(new_pk != g_identity(), EIdentityPublicKey);
that we might miss

move/sources/events.move
```diff
@@ -51,6 +70,8 @@ public struct TransferEvent<phantom T> has copy, drop {
     receiver: address,
     receiver_pk: Element<G>,
     encrypted_amount_receiver: EncryptedAmount,
+    auditor_handles: Option<U32LimbHandles>,
```
benr-ml:
```suggestion
    auditor_decryption_handles: Option<U32LimbHandles>,
```

move/sources/contra.move
```diff
@@ -385,127 +403,99 @@ public fun set_accepts_encrypted_deposits<T>(
...
     );
-    account[TokenAccountKey<T>()].accepts_deposits = true;
 }
 
-/// Re-key the active balance to `new_pk`, aborting if any proof or precondition fails.
-public(package) fun set_public_key_internal<T>(
+/// Like `rekey_token` but soft-fails instead of aborting if the re-key proof does not verify (e.g. a
+/// deposit raced the caller's read). In that case, it instead emits `TryTokenRekeyFailedEvent` and returns `false`, leaving
+/// the token unchanged for a retry. Still aborts on an identity `new_pk` or unmerged pending deposits.
+public fun try_rekey_token<T>(
```
benr-ml: previously it also unpaused

ts-sdk/src/auditor.ts
```diff
@@ -1,144 +1,68 @@
...
-import type {
-	AuditorVersionEntry,
-	ContraAuditorOptions,
-	ContraCompatibleClient,
-	ContraPackageConfig,
-	VerifiedKeyEncryption,
-} from './types.js';
+import type { ContraAuditorOptions } from './types.js';
+
+/** `2^16`, the base regrouping two u16 limbs into one u32 auditor limb. */
```
benr-ml: remove

ts-sdk/src/auditor.ts
```diff
@@ -1,144 +1,68 @@
...
+const SHIFT_16 = 1n << 16n;
+/** `2^32`, the base combining the two u32 auditor limbs into the u64 amount. */
```
benr-ml: remove

ts-sdk/src/auditor.ts
```diff
@@ -1,144 +1,68 @@
...
- * UpdatedPublicKeyEvent events.
+ * The two u32-limb commitments are regrouped from the receiver limbs on the fly
+ * (`C_0 + 2^16 C_1`, `C_2 + 2^16 C_3`), mirroring on-chain `encrypted_amount::ciphertexts_as_u32_limbs`,
+ * and paired with the matching handle to form a twisted ElGamal ciphertext the auditor decrypts.
  */
 export class ContraAuditor {
-	#suiClient: ContraCompatibleClient;
-	#packageConfig: ContraPackageConfig; // Will be static per network in the future.
 	#tokenType: string;
+	#privateKey: PrivateKey;
```
benr-ml: it should handle multiple keys (if auditor key is rotated)
match based on the event's auditor_pk

move/sources/contra.move
jonas-lj: We don't - this is left-over from an earlier iteration.

move/sources/contra.move
jonas-lj: True. We should pause per token when rekeying and the unpause on a succesful rekey. The user can then retry for the tokens where the rekey failed.

move/sources/contra.move
jonas-lj: Good idea!

move/sources/contra.move
jonas-lj: I think it's cleaner to name them after what they are and not what they're used for. Also, it's only used internally. The transfer functions takes an AuditorPackage.

jonas-lj: Well, internally and exposed in events.

move/sources/contra.move
jonas-lj: Sure. Since this is just for public keys here, this could be called PublicKey instead. This also makes the type arguments a bit cleaner and the error, EIdentityPublicKey, still informative.

jonas-lj: Done

ts-sdk/src/auditor.ts
```diff
@@ -1,144 +1,100 @@
...
-		};
-		const privateKey = this.recoverPrivateKey(verified, pointFromBcs(parsed.pk));
-		return new TokenAccount(address, this.#tokenType, this.#packageConfig, privateKey);
+	decryptTransferAmount(
+		encryptedAmountReceiver: EncryptedAmount,
+		auditorHandles: readonly RistrettoPoint[],
+		auditorPk: RistrettoPoint,
+	): bigint {
+		if (auditorHandles.length !== 2) {
+			throw new Error(
```
benr-ml: the errors here should be typed (they are not internal)

ts-sdk/src/auditor.ts
```diff
@@ -1,144 +1,100 @@
...
-		const verified = {
-			ciphertext: parsed.verified_key_encryption.ciphertext.map((raw) =>
-				MultiRecipientEncryption.fromBcs(raw),
-			),
-			version: parsed.verified_key_encryption.version,
-		};
-		const privateKey = this.recoverPrivateKey(verified, pointFromBcs(parsed.pk));
-		return new TokenAccount(address, this.#tokenType, this.#packageConfig, privateKey);
+	decryptTransferAmount(
+		encryptedAmountReceiver: EncryptedAmount,
```
benr-ml: the previous function could serve different events, thus it used low level encryption, etc
however now this function will only be used with TransferEvent, no? if so, let's make this function work directly with TransferEvent

ts-sdk/src/client.ts
```diff
@@ -158,20 +168,61 @@ export class ContraClient {
 			include: { content: true },
 		});
 
+		// For auto-registration, resolve any missing token accounts from the receiver's `Account`
```
benr-ml: this is non trivial logic and i'm not sure we want to support this "auto generate in the same ptb" in the sdk (we won't use it in free-tier since it creates objs; there is a race condition with other transactions that may create the same dest account and fail my tx).
for now i think we can say that the wallet will identify that the dest account is not initialized, and ask the user to init it explicitly in one tx, and then transfer

ts-sdk/src/client.ts
```diff
@@ -286,35 +344,96 @@ export class ContraClient {
...
+			return { tokenType: types[i], registered: true, publicKey };
+		});
+	}
+
+	/**
+	 * Enumerate the token types registered on `address`'s account by listing its dynamic fields (an
+	 * `Account` stores only `TokenAccount<T>`s as dynamic fields, so the field `valueType`s carry every
+	 * `T`). Returns `[]` when the account has no token accounts (or does not exist).
+	 */
+	async #listTokenTypes(address: string): Promise<string[]> {
```
benr-ml: i dont think we need this - a wallet will have a whitelist of tokens it supports

ts-sdk/src/client.ts
```diff
@@ -286,35 +344,96 @@ export class ContraClient {
...
-	 *   await contraClient.getAuditors('0x2::sui::SUI');
+	 * const { currentPk, previousPk, previousExpirationEpoch } =
+	 *   await contraClient.getAuditor('0x2::sui::SUI');
 	 * ```
 	 *
 	 * Throws the underlying fetch error if any.
 	 */
-	async getAuditors(tokenType: string): Promise<TokenAuditors> {
-		const { auditors } = await this.#getConfidentialToken(tokenType);
+	async getAuditor(tokenType: string): Promise<TokenAuditor> {
```
benr-ml, line 345: do we need to return anything but currentPk?

ts-sdk/src/client.ts
```diff
@@ -406,54 +501,65 @@ export class ContraClient {
...
+					},
+				}),
+			);
+	}
+
+	/**
+	 * Register a `TokenAccount<T>` for `receiver` on their behalf, without any `Auth` — the
+	 * permissionless counterpart to `register`. Only succeeds when the token leaves registration
+	 * permissionless, and `receiver` must already have an `Account`. `transferBatch` and `wrap` call
+	 * this automatically for an unregistered receiver, so use it directly only to pre-register.
```
benr-ml: as above, not sure we want this

ts-sdk/src/client.ts
```diff
@@ -464,31 +570,50 @@ export class ContraClient {
...
+	 * SDK-thrown:
+	 * - `TokenAccountDoesNotExistError` — `receiver` has no `Account` at all.
+	 *
 	 * On-chain aborts:
 	 * - `EAuthorizationError` — invalid `auth`.
 	 * - `ETransferDenied` — the token is paused, the deny list is globally frozen, the receiver
 	 *   is on the deny list, or the receiver's per-account freeze is active.
-	 * - `sui::dynamic_field::EFieldDoesNotExist` — `receiver` is not registered for the token.
+	 * - `ERegistrationNotPermissionless` — `receiver` had no `TokenAccount<T>` and the token's
+	 *   registration is permissioned, so the auto-inserted `register_with_default_pk` aborts.
```
benr-ml: same here

ts-sdk/src/error.ts
```diff
@@ -74,6 +74,15 @@ export class TokenAccountDoesNotExistError extends ContraError {
 	}
 }
 
+/** An `Account` object does not exist on chain for the given owner. */
+export class AccountDoesNotExistError extends ContraError {
+	readonly address: string;
+	constructor(address: string, cause: string) {
```
benr-ml: "Account does not exist" - what can be the cause here?

benr-ml (commented):
Move looks good, now focusing on ts sdk

ts-sdk/src/error.ts
jonas-lj: It was used in getTokenTypes, but since that's removed, we don't need it anymore.

ts-sdk/src/auditor.ts
```diff
@@ -1,18 +1,29 @@
...
 import type { ContraAuditorOptions } from './types.js';
 
 const SHIFT_16 = 1n << 16n;
 const SHIFT_32 = 1n << 32n;
 
+/**
+ * The decoded `TransferEvent` fields an auditor reads (`events.ts` `TransferEvent`, via
+ * `TransferEventBcs.parse`). A full decoded event is structurally assignable.
+ */
+export type DecodedTransferEvent = {
```
benr-ml: hmm, don't we already have the type from the auto generated code?

ts-sdk/src/client.ts
```diff
@@ -464,31 +533,52 @@ export class ContraClient {
...
-	wrap({ coin, receiver, tokenType, memo }: WrapOptions) {
-		return (tx: Transaction): TransactionResult =>
-			tx.add(
+	async wrap({
+		coin,
+		receiver,
+		tokenType,
+		memo,
+	}: WrapOptions): Promise<(tx: Transaction) => TransactionResult> {
+		// Wrapping no longer auto-registers on chain. If the receiver has an `Account` but no
```
benr-ml: remove?

ts-sdk/src/client.ts
```diff
@@ -144,10 +146,17 @@ export class ContraClient {
...
+	 * With `autoRegisterFallback`, a receiver that has no `TokenAccount<T>` yet does not throw: its
+	 * state is taken from the receiver's `Account` (`pk = Account.default_pk`, deposits accepted, not frozen)
+	 * and flagged `needsRegistration`, so the caller can prepend a `register_with_default_pk` call
+	 * before depositing (deposits no longer auto-register on chain). A receiver with no `Account` at
+	 * all still throws.
 	 */
 	async #getAccountStates(
 		addresses: readonly string[],
 		tokenType: string,
+		{ autoRegisterFallback = false }: { autoRegisterFallback?: boolean } = {},
```
benr-ml: remove?

ts-sdk/src/client.ts
```diff
@@ -758,181 +850,271 @@ export class ContraClient {
...
+	 * @example
+	 * ```ts
+	 * const rekey = await client.contra.tryRekeyTokenAccounts({
+	 *   rotations: [{ tokenAccount, newTokenAccount }],
+	 * });
+	 * const tx = new Transaction();
+	 * rekey(tx);
+	 * ```
+	 */
+	async tryRekeyTokenAccounts({
```
benr-ml: do we need it? can't the caller create a ptb with tryRekeyTokenAccount?

ts-sdk/src/client.ts
```diff
@@ -1057,6 +1242,7 @@ export class ContraClient {
 		const receiverStates = await this.#getAccountStates(
 			recipients.map((r) => r.receiverAddress),
 			tokenType,
+			{ autoRegisterFallback: true },
```
benr-ml: remove?

move/sources/auditors.move
```diff
@@ -4,215 +4,163 @@
...
+/// — every transfer carries auditor-readable ciphertexts of the amount, one set of decryption handles
+/// per auditor key.
 ///
-/// The `version` number is incremented on every `update` so that `VerifiedKeyEncryption` values
-/// stored on user accounts can be checked for staleness. `recommended_min_version` is the issuer's
-/// advertised minimum `VerifiedKeyEncryption.version`; it is not enforced on chain. Wallets and
-/// other clients should treat any account whose `VerifiedKeyEncryption.version` is below it as
-/// stale and prompt the user to rotate before transferring.
+/// Empty `current_pks` means auditing is disabled going forward. In steady state `current_pks ==
+/// previous_pks`. A change is expressed by pointing `current_pks` at the new set while `previous_pks`
```
benr-ml: hmm, not previous_pks = []?

move/sources/nizk.move
```diff
@@ -46,20 +40,14 @@ public struct ElGamalProof has drop {
...
-public struct KeyConsistencyProof has drop {
-    a1: vector<Element<G>>, // 8*m elements: a_i * pk_j for all (i, j)
-    a2: vector<Element<G>>, // 8 elements: a_i * g + b_i * h
-    a3: Element<G>, // Single aggregate mask (\sum_i b_i * 2^{32i}) * g.
-    z1: vector<Element<Scalar>>, // 8 scalars: a_i + c * r_i
-    z2: vector<Element<Scalar>>, // 8 scalars: b_i + c * u_i
+/// Proves that a shared set of ElGamal commitments `C_k = r_k*g + m_k*h` is consistently encrypted
+/// under several public keys `pk_i` — every handle `D_{i,k} = r_k*pk_i` reuses `C_k`'s `(r_k, m_k)`.
+/// Folded to `m + 1` points + `2` scalars for `m` keys, versus `m` separate `ElGamalProof`s.
+public struct MultiKeyElGamalProof has drop {
```
benr-ml: why elgamal?

ts-sdk/src/auditor.ts
```diff
@@ -1,144 +1,109 @@
...
-		return new TokenAccount(address, this.#tokenType, this.#packageConfig, privateKey);
+	decryptTransferAmount(event: DecodedTransferEvent): bigint | null {
+		const auditorPks = event.auditor_pks;
+		// No auditor data attached (auditing was disabled for this transfer).
+		if (auditorPks.length === 0) return null;
+		// Find the auditor key this instance holds; its handles sit at the same index in
+		// `auditor_decryption_handles`.
+		let index = -1;
+		let privateKey: PrivateKey | undefined;
+		for (const [i, pkBcs] of auditorPks.entries()) {
```
benr-ml: why not use a map instead of iterating this vector every decrypt?

ts-sdk/src/auditor.ts
```diff
@@ -1,144 +1,109 @@
...
+			if (held) {
+				index = i;
+				privateKey = held.privateKey;
+				break;
+			}
+		}
+		if (privateKey === undefined) {
+			throw new AuditorKeyNotHeldError(pointFromBcs(auditorPks[0].element));
+		}
+		const pair = event.auditor_decryption_handles[index];
```
benr-ml: for now index = 0, no?

ts-sdk/src/client.ts
benr-ml: why do we need this?

ts-sdk/src/client.ts
```diff
@@ -464,31 +534,55 @@ export class ContraClient {
...
 	 */
-	wrap({ coin, receiver, tokenType, memo }: WrapOptions) {
-		return (tx: Transaction): TransactionResult =>
-			tx.add(
+	async wrap({
+		coin,
+		receiver,
+		tokenType,
+		memo,
+		registerReceiver = false,
```
benr-ml: let's remove all auto register

move/sources/auditors.move
jonas-lj: Auditing is disabled if any of current_pks and previous_pks are empty, so this was just meant as a suggestion to the caller for what to do when disabling. Could be clearer though

move/sources/nizk.move
jonas-lj: updated to batched DDH proofs

move/sources/auditors.move
```diff
@@ -4,215 +4,153 @@
...
-/// produced against, so callers can compare it against `Auditors.recommended_min_version` to
-/// detect encryptions that the issuer considers stale.
-///
-/// An empty `ciphertext` means the user's account was registered while the token had no
-/// auditors set.
-public struct VerifiedKeyEncryption has copy, drop, store {
-    ciphertext: vector<MultiRecipientEncryption>,
-    version: u32,
-}
+const EAuditorProofFailed: u64 = 1;
```
benr-ml: should start with 0

move/sources/contra.move
```diff
@@ -93,10 +99,11 @@ const EAllAmountsMustBeUsed: u64 = 6;
 const EAmountsEqualityProofFailed: u64 = 7;
 const EEmptyTransferBatch: u64 = 8;
 const ETooManyReceivers: u64 = 9;
-/// Recovery: transfer or update active balance.
 const EBalancesFull: u64 = 10;
-const EIdentityPublicKey: u64 = 11;
 const EBatchTooLarge: u64 = 12;
+const EReceiverNotRegistered: u64 = 16;
```
benr-ml: update numebrs

move/sources/contra.move
```diff
@@ -112,10 +119,9 @@ const PERMISSIONED_UNWRAP: u8 = 2;
 /// Protocol-id `100` is also reserved by the ts-sdk for `PROTOCOL_VERIFIED_DEC`
 const DST_DDH: u8 = 0x01;
 const DST_ELGAMAL: u8 = 0x02;
-const DST_KEY_CONSISTENCY: u8 = 0x03;
 const DST_RANGE_PROOF_16: u8 = 0x04;
-const DST_RANGE_PROOF_32: u8 = 0x05;
 const DST_BATCH_DDH: u8 = 0x06;
+const DST_AUDITOR_ELGAMAL: u8 = 0x07;
```
benr-ml:
```suggestion
const DST_AUDITOR_DDH: u8 = 0x07;
```

move/sources/decode.move
```diff
@@ -6,23 +6,34 @@
...
 }
 
+/// Build one `PublicKey` per point-encoded part; each is validated non-identity by `public_key`.
+public fun public_keys(parts: vector<vector<u8>>): vector<PublicKey> {
+    parts.map!(|b| public_key(g_from_bytes(&b)))
+}
+
+/// Build one `[lo, hi]` handle pair per consecutive pair of point-encoded `parts` (two u32-limb
+/// handles per transferred amount, flattened in amount order). Aborts if `parts` has an odd length.
+public fun decryption_handles(parts: vector<vector<u8>>): vector<vector<Element<G>>> {
```
benr-ml:
```suggestion
public fun auditor_decryption_handles(parts: vector<vector<u8>>): vector<vector<Element<G>>> {
```

move/sources/events.move
```diff
@@ -36,21 +48,30 @@ public struct WrapEvent<phantom T> has copy, drop {
...
     sender: address,
-    sender_pk: Element<G>,
+    sender_pk: PublicKey,
     seed_point: Element<G>,
     batch_index: u8,
     receiver: address,
-    receiver_pk: Element<G>,
-    encrypted_amount_receiver: EncryptedAmount,
+    receiver_pk: PublicKey,
+    encrypted_amount_receiver: vector<Encryption>,
```
benr-ml: why not EncryptedAmount?

move/sources/events.move
```diff
@@ -36,21 +48,30 @@ public struct WrapEvent<phantom T> has copy, drop {
...
+    auditor_decryption_handles: vector<vector<Element<G>>>,
```
benr-ml: one level, no?

move/sources/events.move
```diff
@@ -36,21 +48,30 @@ public struct WrapEvent<phantom T> has copy, drop {
...
+    auditor_pks: vector<PublicKey>,
```
benr-ml: one pk

move/sources/auditors.move
```diff
@@ -4,215 +4,153 @@
...
+    handles: vector<vector<Element<G>>>,
+    proofs: vector<DdhProof>,
 }
 
-public(package) fun key_version(verified_key_encryption: &VerifiedKeyEncryption): u32 {
-    verified_key_encryption.version
+/// The verified auditor decryption handles for a batch: `handles[r]` is receiver `r`'s `[lo, hi]`
+/// pair, tagged with the auditor's `pk`. Held in `TransferBatch` state (as an `Option`, `none` when
+/// auditing is disabled); `per_receiver` reads one receiver's pair for that receiver's `TransferEvent`.
+public struct VerifiedAuditorHandles has copy, drop, store {
```
benr-ml: - do we need copy?
- what about exposing next() instead of per_receiver() and destroy() instead of drop

move/sources/events.move
jonas-lj: because it's now just two u32 limbs instead of four u16 limbs

benr-ml: hmm, why is the receiver's encrypted amount u32?

move/sources/auditors.move
jonas-lj: I've updated with a next and destroy function and removed copy and drop.

move/sources/events.move
jonas-lj: Why not? It's still decryptable. And it's two group elements fewer per event

benr-ml: it makes decryption slower in wallets while we have the information in move already. does the event size matter here?

move/sources/auditors.move
```diff
@@ -4,215 +4,167 @@
...
+
+/// Consume the auditor data (it has no `drop`, so this is the only way to discard it). In the normal
+/// flow every receiver's pair has been popped by `next` and only the key is left; on a failed batch
+/// it is discarded unpopped. Either way its contents are dropped.
+public(package) fun destroy(auditor_data: Option<VerifiedAuditorHandles>) {
+    if (auditor_data.is_none()) {
+        auditor_data.destroy_none();
+        return
+    };
+    let VerifiedAuditorHandles { handles: _, pk: _ } = auditor_data.destroy_some();
```
benr-ml, line 136: assert they handles is empty

ts-sdk/src/client.ts
```diff
@@ -464,31 +490,41 @@ export class ContraClient {
...
-	wrap({ coin, receiver, tokenType, memo }: WrapOptions) {
-		return (tx: Transaction): TransactionResult =>
-			tx.add(
+	async wrap({
+		coin,
+		receiver,
+		tokenType,
+		memo,
+	}: WrapOptions): Promise<(tx: Transaction) => TransactionResult> {
+		// Deposits never auto-register: fail early if the receiver has no token account.
```
benr-ml, line 524: remove?

ts-sdk/src/types.ts
```diff
@@ -283,20 +272,21 @@ export interface RegisterOptions {
...
 	 */
 	auth?: AuthThunk;
 }
 
+/** Arguments to `ContraClient.tryRegisterWithDefaultPk`. */
+export interface RegisterWithDefaultPkOptions {
+	/** The owner address of the account to register a token account for. Must already have an `Account`. */
+	receiver: string;
+	/** The fully-qualified Move type of the token to register, e.g. `0x2::sui::SUI`. */
+	tokenType: string;
```
benr-ml, line 287: let's be consistent with the order - tokenType always the first

AUDITORS.md
```diff
@@ -119,12 +119,20 @@ The two options trade cost against operational complexity. The table below summa
...
+We implement **Option 1 (per-transfer)** with a **single auditor key per token**. Each transfer attaches an auditor-readable copy of the amount under that key; the auditor never learns a user's viewing key, so balances stay encrypted only under the user's own key and users may reuse one account public key across tokens.
 
-- **Routine rotation.** Wallets rotate user viewing keys roughly every month.
-- **Triggered rotation.** When the issuer signals on-chain that rotation is recommended (expected only in response to key leakage), wallets rotate immediately.
-- **Inactive accounts.** Inactive users may not rotate until they are next active, and would still receive deposits encrypted under the old key in the meantime. To limit exposure, wallets are encouraged not to send to accounts that haven't rotated within a month after an issuer-signaled rotation.
+Concretely (see `auditors.move` / `contra.move` / `nizk.move`):
```
benr-ml: let's remove everything from l124

README.md
```diff
@@ -134,6 +134,12 @@ The TS-SDK's `transfer` and `unwrap` methods default to `merge: true`, which pre
 
 In the failure case, a second attempt with `merge: false` will succeed immediately, because the previously-pending deposits have already been folded into the active balance by the first transaction's `merge` call, so the proof now only needs to match the active balance and is unaffected by any further deposits that may have arrived in the meantime.
 
+### Key rotation
+
+Each token's balance is encrypted under its own `TokenAccount.pk`, chosen at `register` and rotated **independently** with `rekeyTokenAccount(newTokenAccount)` — the new key is explicit and unrelated to any other token or to the account's key. Re-keying is **lazy and O(1)** per token: it touches only that token's active balance (`tryRekeyTokenAccounts` is a convenience that merges + re-keys one or more tokens — each `(current, new)` pair to its own new key — in a single, optimistic (soft-failing) PTB). A not-yet-re-keyed token is a normal, decryptable state — deposits keep landing under its current key until it is re-keyed. Separately, `Account.default_pk` is an **optional** default key that only drives `register_with_default_pk`; `setDefaultPkAsSender` sets or clears it (O(1), touches no balance).
```
benr-ml: remove from "Re-keying" until the end of the parag

README.md
```diff
@@ -134,6 +134,12 @@ The TS-SDK's `transfer` and `unwrap` methods default to `merge: true`, which pre
...
+
+> **Retain the old key until every token is re-keyed.** A not-yet-re-keyed token's balance is still encrypted under the _old_ key, and both re-keying it (the proof witness is `newSk · oldSk⁻¹`) and decrypting it require that old key. Discarding the old private key while any token is still under it makes that token's balance permanently unrecoverable. Use `ContraClient.getTokenKeys(address)` to list every token's current key (pass specific `tokenTypes` to narrow it) — an old key is safe to delete only once no token still reports it, and after a `tryRekeyTokenAccounts` any token that still reports its old key is one whose re-key soft-failed and needs a retry.
```
benr-ml: let's make it less verbose

README.md
```diff
@@ -208,23 +214,21 @@ By default `register`, `wrap`, and `unwrap` are open to any holder of `T`. The i
...
-See [Auditor Support in Confidential Transfers](AUDITORS.md) for more details on the per-transaction and per-account auditing.
+The contract implements **per-transfer auditing** with a single auditor key per token (rotatable, with a grace window): the sender attaches an auditor-readable copy of each transfer amount (two u32-limb decryption handles per receiver) derived from the same range-proven commitments the receiver gets, plus one short zero-knowledge proof per u32 limb (two per receiver) that ties the auditor's handle to the receiver's own range-proven commitment. Balances stay encrypted only under the user's own key, so the same account key can be reused across tokens. See [Auditor Support in Confidential Transfers](AUDITORS.md) for the design and the alternatives considered.
```
benr-ml:
```suggestion
The contract implements **per-transfer auditing** with a single auditor key per token: the sender attaches an auditor-readable copy of each transfer amount. Balances stay encrypted under the user's own key. See [Auditor Support in Confidential Transfers](AUDITORS.md) for the design and the alternatives considered.
```

README.md
```diff
@@ -208,23 +214,21 @@ By default `register`, `wrap`, and `unwrap` are open to any holder of `T`. The i
...
-3. **Collect historical secrets.** The auditor assembles a `Map<version, (index, secretKey)>` covering every on-chain version it wants to be able to decrypt for. For versions that the auditor itself generated, the secret is its own. For older versions (set up before this auditor existed), the issuer or prior auditors must hand over the corresponding secrets — otherwise accounts that registered against those older versions remain opaque to this auditor.
-4. **Initialize [`ContraAuditor`](ts-sdk/src/auditor.ts).** Construct the SDK with that version map plus a precomputed discrete-log table (the larger `numBits` is, the faster decryption will be).
+1. **Generate an auditor keypair off-chain.** Each auditor generates a Twisted ElGamal keypair `(sk, pk)` over Ristretto255 and hands the **public key** to the token issuer.
```
benr-ml:
```suggestion
1. **Generate an auditor keypair off-chain.** The token issuer generates a Twisted ElGamal keypair `(sk, pk)` over Ristretto255.
```

README.md
```diff
@@ -208,23 +214,21 @@ By default `register`, `wrap`, and `unwrap` are open to any holder of `T`. The i
...
+2. **Issuer sets the on-chain auditor keys.** The issuer calls [`update_auditors`](move/sources/contra.move) with two key vectors: `current_pks` (the keys tried first on a transfer; empty disables auditing going forward) and `previous_pks` (also accepted, for a grace window). The two need not be the same length. No per-user escrow. Grace is caller-driven, with no expiration epoch: to change the set with a grace window set `current_pks` to the new keys (which may be fewer, or empty to disable) while `previous_pks` still holds the outgoing keys, so in-flight transfers built against the old keys still verify; end the grace by setting `previous_pks` equal to `current_pks`.
```
benr-ml:
```suggestion
2. **Issuer sets the on-chain auditor keys.** The issuer calls [`update_auditors`](move/sources/contra.move) with two key vectors: `current_pks` (the keys tried first on a transfer; empty disables auditing going forward) and `previous_pks` (also accepted, for a grace window). The two need not be the same length. Grace is caller-driven: to change the set with a grace window set `current_pks` to the new keys while `previous_pks` still holds the outgoing keys, so in-flight transfers built against the old keys still verify; end the grace by setting `previous_pks` equal to `current_pks`.
```

README.md
```diff
@@ -208,23 +214,21 @@ By default `register`, `wrap`, and `unwrap` are open to any holder of `T`. The i
...
 
-Once initialized, the auditor can call `getTokenAccount(address)` on the SDK. It fetches the on-chain `TokenAccount<T>`, decrypts the user's secret key, and returns a fully-keyed `TokenAccount` object. That object can then be passed to `ContraClient.getBalance` (active and pending balances) or to `EncryptedAmount.decrypt` for amounts that appear in event payloads encrypted to that user.
+Auditing is per transfer, not per account. For each [`TransferEvent`](move/sources/events.move), the auditor calls `decryptTransferAmount(event)`: it uses the held key that matches the event's `auditor_pk` (`none` when auditing is disabled), reads that auditor's two handles from `auditor_decryption_handles`, takes the two u32-limb commitments straight from the event's `encrypted_amount_receiver`, pairs each with the matching handle, and BSGS-decrypts to recover that transfer's amount. A token has one auditor key at a time, but an auditor can hold several keys, so keeping the rotated-out keys lets it read transfers made both before and after a key rotation. The auditor sees each transfer's amount, sender, and receiver — but never a user's viewing key or account balance.
```
benr-ml:
```suggestion
Auditing is per transfer. For each [`TransferEvent`](move/sources/events.move), the auditor calls `decryptTransferAmount(event)` which decrypts and recovers that transfer's amount. A token has one auditor key at a time, but an auditor can hold several keys, so keeping the rotated-out keys lets it read transfers made both before and after a key rotation.
```

move/sources/events.move
jonas-lj: Maybe not that much. The cost of emitting an event depends on the size, but reducing to two limbs is not free on-chain (and also not on the wallet). I'll revert.

move/sources/auditors.move
jonas-lj: We can't do that here since we need to destroy the handles in case the optimistic transfer fails. I'll add a destroy_empty next to this so it's clear when we assert.

benr-ml (approved):

PR #25, opened 2026-08-12
[move] add guardian interfaces

joyqvq:


move/sources/guardian.move
```diff
@@ -0,0 +1,399 @@
...
+}
+
+// Issuer operations (ManagementCap-gated by the wrappers in contra.move).
+
+/// Each `Some` field is applied, `None` is a no-op; `pcrs` applies first, bumping
+/// `version` without invalidating existing keys. Raising `min_version` revokes every
+/// key stamped below it.
+public(package) fun update(
+    self: &mut GuardianPolicy,
+    pcrs: Option<Pcrs>,
```
benr-ml: can remove the optionals, always set all

move/sources/guardian.move
```diff
@@ -0,0 +1,399 @@
...
+    min_version: Option<u16>,
```
benr-ml: instead of storing min_version, why not remove all the registerd enclaves that are below min_version?

move/sources/guardian.move
```diff
@@ -0,0 +1,399 @@
...
+    url: String,
+    /// Incremented per PCR change.
+    version: u16,
+    /// A key is valid iff `key.version >= min_version`; raising this invalidates
+    /// old-image keys.
+    min_version: u16,
+    /// The expected enclave image measurements.
+    pcrs: Pcrs,
+    /// The fleet's registered keys — one per enclave instance, keyed by `signing_pk`.
+    guardian_enclave_keys: VecMap<Ed25519PublicKey, GuardianEnclaveKey>,
```
benr-ml: nit - instead of using pk as the key (sent in every tx), use VecMap<u8, GuardianEnclaveKey>
- use indexes 0..MAX_GUARDIAN_ENCLAVE_KEYS
- when registering a new enclave, find an index that is not in use

PR #26, opened 2026-08-13
auditors: batch the auditor handle proof into one folded ElGamal proof

jonas-lj:
Per-transfer auditing attached one `nizk::DdhProof` per (receiver, u32-limb) per batch. This replaces them with a single `ElGamalProof` over all `2N` auditor ciphertexts, at constant size.


move/sources/auditors.move
```diff
@@ -136,32 +132,26 @@ public(package) fun destroy(auditor_data: Option<VerifiedAuditorHandles>) {
...
-    vector::tabulate!(n, |r| {
-        let receiver_handles = receiver_amounts[r].handles_u32();
-        let bases = vector[*receiver_amounts[r].pk().as_element(), auditor_pk];
-        vector::tabulate!(U32_LIMBS, |l| {
-            let images = vector[receiver_handles[l], decryption_handles[r][l]];
-            proofs[r * U32_LIMBS + l].verify_ddh(dst, &bases, &images)
-        }).all!(|ok| *ok)
-    }).all!(|ok| *ok)
+    if (pks.length() != 1 || handles.length() != n) return false;
+    let mut encryptions = vector<Encryption>[];
```
benr-ml, line 149: nit - during grace period, this calculation is done twice instead of once

benr-ml (approved):

move/sources/auditors.move
jonas-lj: Will be fixed in a follow up

PR #33, opened 2026-08-14
Move: consistent/well-formed encryption types + decoupled transfer verification

jonas-lj:


move/sources/contra.move
```diff
@@ -519,34 +530,113 @@ public fun wrap<T>(
...
+/// transferred value re-encrypted under `receiver_pks[i]` and proven consistent by
+/// `receiver_consistency_proofs[i]`. `sender_consistency_proof` folds the `new_balance` limbs and the
+/// total (under `sender`'s key) into one proof, `range_proofs` range-check every limb, and
+/// `total_sender_handle` is the single sender-keyed decryption handle for the transfer total (its
+/// commitment is reconstructed from the receiver amounts).
+public fun prepare_transfer_amounts<T>(
+    sender: &Account,
+    receiver_pks: vector<PublicKey>,
+    receiver_amounts: vector<EncryptedAmount>,
+    receiver_consistency_proofs: vector<ElGamalProof>,
```
benr-ml: consistency proof? the elgamal are pok

move/sources/contra.move
```diff
@@ -519,34 +530,113 @@ public fun wrap<T>(
...
+    new_balance: EncryptedAmount,
+    sender_consistency_proof: ElGamalProof,
```
benr-ml: transfer_amount_pok?

move/sources/contra.move
```diff
@@ -519,34 +530,113 @@ public fun wrap<T>(
...
+    range_proofs: RangeProofs,
+    total_sender_handle: Element<G>,
+): (vector<WellFormedEncryptedAmount>, WellFormedEncryptedAmount, ConsistentEncryption) {
```
benr-ml: ConsistentEncryption abstraction is used only to pass information between prepare_transfer_amounts and batch_transfer, and those functions must be called together - why do we need this extra abstraction, instead of doing that in batch_transfer? splitting that logic introduces extra checks and could be error-prone

benr-ml: if we already have a special case for the sender's encs, we can have it inside transfer or even in batch_into_well_formed. total_sender below is actually WFE as well (as we check its ElGamal directly, and the range proof indirectly since we check that for receiver_amounts)

jonas-lj: > if we already have a special case for the sender's encs, we can have it inside transfer or even in batch_into_well_formed. total_sender below is actually WFE as well (as we check its ElGamal directly, and the range proof indirectly since we check that for receiver_amounts)

The total_sender is just a singe encryption with no limbs (since we don't need them). And even if we considered the limbs, it's the sum of the receiver_amounts, which are well-formed, but their sum doesn't have to be.

jonas-lj: > ConsistentEncryption abstraction is used only to pass information between prepare_transfer_amounts and batch_transfer, and those functions must be called together - why do we need this extra abstraction, instead of doing that in batch_transfer? splitting that logic introduces extra checks and could be error-prone

I added it to simplify the batched_transfer function. The computation done inside `prepare_transfer_amounts` is pretty non-trivial, but I agree that it requires an extra step of validation (that all verification was done for the right DST and public key) which is not ideal. But adding a handful of arguments + all this combinatorics to create the right proof batches is not simpler imo.

move/sources/contra.move
jonas-lj: hmm, but it's for both the new balance and the total_sender. Maybe `sender_encs_pok` or something like that?

move/sources/contra.move
jonas-lj: True, that's more accurate. "consistency" was used to indicate that it's proof that the encryption was correctly formed

move/sources/contra.move
jonas-lj: We can make prepare_transfer_amounts a private function called from batched_transfer. I think this would be the cleanest option

move/sources/encrypted_amount.move
```diff
@@ -24,16 +24,17 @@ const LIMB_BITS: u8 = 16;
...
 const U16_LIMBS: u64 = 4;
 
-/// Maximum number of amounts covered by a single Bulletproof chunk.
+/// Maximum number of encryptions covered by a single Bulletproof chunk.
 /// `sui::rangeproofs::verify_bulletproofs_with_dst_ristretto255` caps the aggregated commitment
-/// count at 32 for `LIMB_BITS = 16`, and each amount contributes 4 limb commitments, so a single proof
-/// covers at most `32 / 4 = 8` amounts.
-const MAX_BATCH_SIZE: u64 = 8;
+/// count at 32 for `LIMB_BITS = 16`, and each encryption contributes one commitment.
+const MAX_BATCH_SIZE: u64 = 32;
```
benr-ml:
```suggestion
const MAX_RANGE_PROOF_BATCH_SIZE: u64 = 32;
```

move/sources/contra.move
```diff
@@ -1025,6 +1036,69 @@ fun dst(session_id: vector<u8>, protocol_id: u8): vector<u8> {
...
+    sender_encs_pok: ElGamalProof,
+    range_proofs: RangeProofs,
+): (VerifiedTransferAmounts, WellFormedEncryptedAmount) {
+    let n = receiver_amounts.length();
+    assert!(receiver_pks.length() == n && receiver_encs_pok.length() == n, EMismatchedBatchLength);
+    let token_account = &sender[TokenAccountKey<T>()];
+
+    // Each receiver amount is proven consistent under its own key. Collect every limb for the
+    // single range batch.
+    let mut range_inputs = vector::tabulate!(n, |i| {
```
benr-ml:
```suggestion
    let mut enc_amounts_as_limbs = vector::tabulate!(n, |i| {
```

move/sources/contra.move
```diff
@@ -1025,6 +1036,69 @@ fun dst(session_id: vector<u8>, protocol_id: u8): vector<u8> {
...
+    // The total's commitment is reconstructed from the receiver commitments and the given handle.
+    let total_sender = twisted_elgamal::new(
+        encrypted_amount::sum_ciphertexts(&receiver_amounts),
+        total_sender_handle,
+    );
+
+    // Sender side: the new-balance limbs and the total are verified under `sender_pk` in one proof.
+    let mut sender_encryptions = new_balance.limbs();
+    sender_encryptions.push_back(total_sender);
+    let mut sender_consistents = encrypted_amount::verify_consistent(
```
benr-ml:
```suggestion
    let mut sender_encs_as_limbs = encrypted_amount::verify_consistent(
```

move/sources/encrypted_amount.move
```diff
@@ -24,16 +24,16 @@ const LIMB_BITS: u8 = 16;
...
```
benr-ml:
```suggestion
const MAX_RANGE_PROOF_BATCH_SIZE: u64 = 32;
```

benr-ml (approved):

PR #36, opened 2026-08-17
[guardian] add wire types and plaintext checks

joyqvq:
focus on
1. sealing.rs: handling of client request encrypted to a list of enclave pks (the symmetric key is encrypted to all keys where the payload is encrypted once with the symmetric key). and unseal them for a particular enclave instance. 
2. checks.rs verifies the openings for encrypted balances and plaintext for old and new. check plaintext for overdraft and overflow.
3. wire types with move matching https://github.com/MystenLabs/confidential-transfers/pull/41

the HTTP entrypoints, enclave server code and scripts for enclave will be in the next pr


guardian/core/src/lib.rs
```diff
@@ -0,0 +1,209 @@
...
+        Self {
+            signing_sk,
+            hpke_sk,
+            keys,
+        }
+    }
+
+    /// The public keys to embed in the attestation's `user_data`, laid out as
+    /// `guardian::parse_user_data` expects.
+    pub fn keys(&self) -> &EnclaveKeys {
```
jonas-lj: "keys" is a bit ambiguous here since all three fields are keys. "enclave_keys" perhaps?

guardian/core/src/types.rs
```diff
@@ -0,0 +1,191 @@
...
+
+impl From<&RistrettoPoint> for MoveElement {
+    fn from(point: &RistrettoPoint) -> Self {
+        Self(point.to_byte_array().to_vec())
+    }
+}
+
+/// Mirrors `contra::twisted_elgamal::Encryption`.
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct MoveCiphertext {
```
jonas-lj: To better align with the Move side
```suggestion
pub struct MoveEncryption {
```

guardian/core/src/checks.rs
```diff
@@ -0,0 +1,521 @@
...
+            if recipients.is_empty() {
+                return Err(GuardianError::EmptyTransfer);
+            }
+
+            let total_txn_amount = recipients
+                .iter()
+                .try_fold(0u64, |acc, r| acc.checked_add(r.amount))
+                .ok_or(GuardianError::Overdraft)?;
+
+            check_balances(
```
jonas-lj: Since this runs first, the error for a wrong total amount doesn't reveal what recipient the wrong amount was for, just as a NewBalanceMismatch

guardian/core/src/checks.rs
```diff
@@ -0,0 +1,521 @@
...
+
+            // Each ciphertext must encrypt its claimed amount to that recipient.
+            for (i, r) in recipients.iter().enumerate() {
+                r.encrypted_amount
+                    .verify(
+                        &RistrettoScalar::from(r.amount),
+                        &r.receiver_pk,
+                        &r.blinding,
+                    )
+                    .map_err(|_| GuardianError::AmountMismatch(i))?;
```
jonas-lj: This returns an error for the first wrong amount. Maybe that's okay but we could also check all amounts and report the wrong ones or at least be specific about what we're reporting here (something like "first amount mismatch is to recipient i")

guardian/core/src/checks.rs
```diff
@@ -0,0 +1,521 @@
...
+            old_balance,
+        } => {
+            if recipients.is_empty() {
+                return Err(GuardianError::EmptyTransfer);
+            }
+
+            let total_txn_amount = recipients
+                .iter()
+                .try_fold(0u64, |acc, r| acc.checked_add(r.amount))
+                .ok_or(GuardianError::Overdraft)?;
```
jonas-lj, line 49: This does mean that the sender is trying to send more token than he has but it's also more tokens than there can exist, so I think this should be a different error type.

guardian/core/src/types.rs
```diff
@@ -0,0 +1,191 @@
...
+
+/// The payload the enclave signs, either a transfer or an unwrap transaction.
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub enum RequestPayload {
+    Transfer {
+        sender_pk: MoveElement,
+        receiver_pks: Vec<MoveElement>,
+        old_encrypted_balance: MoveCiphertext,
+        new_encrypted_balance: MoveCiphertext,
+        encrypted_amounts: Vec<MoveCiphertext>,
```
jonas-lj: Would it make sense to send the encrypted amount (the u64 value as 4xu16 limbs) here s.t. the guardian can also verify that the limbs are in range (are indeed u16s)?

guardian/core/src/sealing.rs
```diff
@@ -0,0 +1,149 @@
...
+    Ok(SealedRequest {
+        version: SEALED_REQUEST_VERSION,
+        sealed: enc_pks
+            .iter()
+            .map(|pk| seal(pk, &plaintext))
+            .collect::<Result<_>>()?,
+    })
+}
+
+pub fn seal(enc_pk: &EncryptionPublicKey, plaintext: &[u8]) -> Result<SealedEnvelope> {
```
jonas-lj: I think the argument here should be an UnsealedRequest, not just a binary plaintext.

jonas-lj:
Can you add some CI for the rust code also?

guardian/core/src/types.rs
joyqvq: @jonas-lj edit: thinking about this more - 

i updated that the client needs to submit in u16 limbs for plaintext and encrypted amounts for request, then the Guardian collapses them on the server side and verifies the resulting amount (the client is responsible for it being canonical). 

i think this is the most efficient and keep the guardian simple. since contra already enforces each transfer limb’s u16 range on chain already. i dont think we need guardian to check that again.

joyqvq:
@jonas-lj i addressed all of your comments and have a follow up for one of them. 

i also made a few additional improvements (other than adding more tests and organizational changes):

- encrypt the request payload once with ChaCha20-Poly1305, then HPKE-wrap only the payload key for each enclave
- now wrapped keys are indexed by encryption public key, so it saves the sequential scan per envelop
- organized verify_binding in the order of cheapest -> expensive checks 

PTAL! 🙏

guardian/core/src/types.rs
jonas-lj: But the guardian already double-checks some things that we check on-chain (the correctness of the balances), so why not also check the range? It'd be simple to do for the guardian (just check that all limbs are < 2^16 for the deposit amounts).

guardian/core/src/move_types.rs
```diff
@@ -0,0 +1,233 @@
...
+            l2.into_ciphertext()?,
+            l3.into_ciphertext()?,
+        ],
+    })
+}
+
+/// Exact BCS mirror of `contra::authority::Binding`. Changing its layout or meaning requires
+/// incrementing the request version and updating the cross-language fixtures.
+#[derive(Debug, Serialize)]
+pub(crate) enum Binding {
```
jonas-lj, line 110: This has no token or account in it so an approval is replayable across two tokens if the
client reuses a pk and its blindings. I think we should add the session id as a field on both variants?

guardian/core/src/lib.rs
```diff
@@ -0,0 +1,218 @@
...
+
+/// Result type returned by Guardian validation and unsealing operations.
+pub type Result<T> = std::result::Result<T, GuardianError>;
+
+/// A request-validation or sealed-transport failure.
+#[derive(Debug, Error, PartialEq, Eq)]
+pub enum GuardianError {
+    #[error("transfer has no recipients")]
+    EmptyTransfer,
+    #[error("transfer has {0} recipients; maximum is 255")]
```
jonas-lj:
```suggestion
    #[error("transfer has {0} recipients; maximum is {MAX_BATCH_RECIPIENTS}")]
```

guardian/core/src/sealing.rs
```diff
@@ -0,0 +1,81 @@
...
+        return Err(GuardianError::InvalidSealedRequest);
+    }
+
+    // Select this enclave's wrapped payload key.
+    let wrapped = request
+        .wrapped_keys
+        .get(&enc_pk)
+        .ok_or(GuardianError::NotARecipient)?;
+    let encapped =
+        EncappedKey::from_bytes(&wrapped.encapped_key).map_err(|_| GuardianError::NotARecipient)?;
```
jonas-lj: This is not the right error. `MalformedEncappedKey`?

guardian/core/src/sealing.rs
```diff
@@ -0,0 +1,81 @@
...
+    // Open the payload key with this enclave's HPKE private key.
+    let payload_key = hpke::single_shot_open::<ChaCha20Poly1305, HkdfSha256, X25519HkdfSha256>(
+        &OpModeR::Base,
+        sk,
+        &encapped,
+        HPKE_INFO,
+        &wrapped.encrypted_key,
+        &[request.version],
+    )
+    .map_err(|_| GuardianError::NotARecipient)?;
```
jonas-lj: Also not the right error. `PayloadKeyUnwrapFailed`?

guardian/core/src/types.rs
```diff
@@ -0,0 +1,178 @@
...
+
+/// An unsealed on-chain operation and the private witnesses needed to validate it.
+#[derive(Serialize, Deserialize)]
+pub enum UnsealedRequest {
+    TransferRequest {
+        old_encrypted_balance: EncryptedAmount,
+        new_encrypted_balance: EncryptedAmount,
+        recipients: Vec<Recipient>,
+        /// Sender secret key used only to verify the balance openings and derive `sender_pk`.
+        x_a: PrivateKey,
```
jonas-lj, line 94: We don't need to give the private key here since the SDK's `decryptWithProof` can prove the old balance
opens to the claimed value. And the new balance doesn't need either since the client picks that randomness and can send per-limb blindings like the recipients do.

guardian/core/src/move_types.rs
joyqvq: i had this discussion with @benr-ml that the guardian should not care about the domain separation. it is the onchain function's job to enforce it

guardian/core/src/types.rs
joyqvq: i also had this discussion with @benr-ml on this. iiuc with decryptWithProof, the old-balance validity depends on a DDH proof right? 

the goal here is to let guardian directly check balance openings. cc @benr-ml what do you think?

guardian/core/src/sealing.rs
joyqvq: this turns out to be in fallible given its [u8; 32] already. revised

guardian/core/src/types.rs
joyqvq: oh i thought you meant to verify the opening per limb instead of collapsing. sure we can check the range - i made the per recipient amount in limbs in u16 instead one u64 amount so the range is checked on wire at request type

guardian/core/src/types.rs
```diff
@@ -0,0 +1,178 @@
...
+        let EncryptionPublicKeyBytes(bytes) = Deserialize::deserialize(deserializer)?;
+        EncryptionPublicKey::from_bytes(&bytes).map_err(serde::de::Error::custom)
+    }
+}
+
+/// One shared encrypted payload and one wrapped payload key per live on-chain enclave key.
+#[derive(Serialize, Deserialize)]
+pub struct SealedRequest {
+    pub(crate) version: u8,
+    pub(crate) payload_nonce: [u8; 12],
```
benr-ml, line 73: what is this?

guardian/core/src/types.rs
benr-ml: yes, the goal is to not depend on the zk proofs, otherwise we do the same checks as we do in Move

guardian/core/src/types.rs
```diff
@@ -0,0 +1,178 @@
...
+    pub(crate) encrypted_payload: Vec<u8>,
+    pub(crate) wrapped_keys: BTreeMap<EncryptionPublicKeyBytes, WrappedPayloadKey>,
```
benr-ml, line 75: i don't follow how the enc works - don't we have 1 symmetric key and n encryption of that key? do we need n encapped_key?

joyqvq: yes, that’s the design: one symmetric payload key and n HPKE encryptions of that key, one per enclave. 

each HPKE encryption produces both an encrypted_key and its recipient-specific encapped_key, so we need n of each.

am i missing something?

guardian/core/src/types.rs
joyqvq: used by: ChaCha20Poly1305(payload_key, payload_nonce, plaintext_request)

guardian/core/src/types.rs
benr-ml: using the lib as a black box results in the following message (roughly):
- ChaCha20Poly1305(k: key, data) where k is a random key
- for i=1..n, KEM(pk_i, s_i), ChaCha20Poly1305(s_i, k), where s_i is a random key

Instead, one could send:
- ChaCha20Poly1305(k: key, data) where k is a random key
- for i=1..n, KEM(pk_i, k)

not super dramatic but a bit clearer and shorter.

guardian/core/src/types.rs
joyqvq: i realized i only addressed the plaintext side: sending the amount as [u16; 4] ensures each plaintext limb is in range, but our current collapsed opening check does not prove that each ciphertext limb encrypts its corresponding plaintext limb.

@benr-ml should we have the client also send the per-limb blindings and verify all four ciphertext openings independently? that would enforce the invariant in the guardian, but requires 4 opening checks instead of 1. 

alternatively, we can keep the single collapsed check and rely on the existing on-chain range proofs for the ciphertext limbs. which one do we prefer?

guardian/core/src/types.rs
joyqvq: i looked at a few crypto libs and looks like their standard APIs always include a fresh ephemeral public key per recipient, so we will need to implement some custom construction from lower-level primitives. 

i guess if the number of keys is low (lets say its capped at 16) its just a relatively small overhead, should we just use `hpke::single_shot_seal` for this to avoid introducing a custom impl? we can move this to fastcrypto, or maybe modify the ECIES impl there?

joyqvq:
aligned move side changes from https://github.com/MystenLabs/confidential-transfers/pull/41 on latest commit

guardian/core/src/types.rs
benr-ml: checking u16 is not critical imp since it's a property required by the cryptographic protocol
for the integrity of balances we only care about the amounts staying consistent

guardian/core/src/types.rs
benr-ml: let's leave a TODO for now

guardian/core/src/types.rs
```diff
@@ -0,0 +1,178 @@
...
+
+impl From<&EncryptionPublicKey> for EncryptionPublicKeyBytes {
+    fn from(pk: &EncryptionPublicKey) -> Self {
+        use hpke::Serializable;
+
+        Self(pk.to_bytes().into())
+    }
+}
+
+/// The enclave's public keys. Their BCS encoding is the 32-byte signing key followed by the
```
benr-ml:
```suggestion
/// The enclave's public keys. 
```

guardian/core/src/types.rs
```diff
@@ -0,0 +1,178 @@
...
+
+impl<'de> Deserialize<'de> for EncryptedAmount {
+    fn deserialize<D: serde::Deserializer<'de>>(deserializer: D) -> Result<Self, D::Error> {
+        crate::move_types::deserialize_encrypted_amount(deserializer)
+    }
+}
+
+/// One transfer recipient, including its on-chain fields and enclave-only plaintext witnesses.
+#[derive(Serialize, Deserialize)]
+pub struct Recipient {
```
benr-ml:
```suggestion
pub struct TransferRecipient {
```

guardian/core/src/lib.rs
```diff
@@ -0,0 +1,223 @@
...
+pub struct EnclaveKeyPair {
+    signing_sk: Ed25519KeyPair,
+    hpke_sk: <X25519HkdfSha256 as Kem>::PrivateKey,
+    enclave_keys: EnclaveKeys,
+}
+
+impl EnclaveKeyPair {
+    /// Generate fresh keys from the operating system RNG at boot.
+    pub fn generate() -> Self {
+        let mut rng = rand::thread_rng();
```
benr-ml: receive from the caller

benr-ml (approved):

PR #38, opened 2026-08-18
[Move] Refactor: Decouple the contra and balance modules better

jonas-lj:
Cleaner decoupling between the contra and balance modules. Refactor only.

`TokenAccount<T>` held a key, two encrypted balances and a public balance, and `contra.move` reached into all four. It now holds one `balance::EncryptedBalance<T>`, and `balance.move` owns everything that touches values like deposits, merges, verification, withdrawals, re-keying — checking each amount against its own key and deriving its own Fiat-Shamir transcripts. `contra.move` keeps authorization, freezing, deny lists, events and pool custody. Adds also a new session module to handle the dst's.


move/sources/session.move
```diff
@@ -0,0 +1,65 @@
...
+
+fun dst(self: &Session, tag: u8): vector<u8> {
+    let mut bytes = self.id;
+    bytes.push_back(tag);
+    bytes
+}
+
+// === Test Helpers ===
+
+#[test_only]
```
benr-ml: do we still need those?

move/sources/session.move
```diff
@@ -0,0 +1,65 @@
...
+const DST_DDH: u8 = 0x01;
+const DST_ELGAMAL: u8 = 0x02;
+const DST_RANGE_PROOF_16: u8 = 0x04;
+const DST_BATCH_DDH: u8 = 0x06;
+const DST_AUDITOR_ELGAMAL: u8 = 0x07;
+
+// === Structs ===
+
+/// A domain separator unique to one account's holdings of one token.
+public struct Session has copy, drop, store {
```
benr-ml:
```suggestion
public struct SessionId has copy, drop, store {
```

move/sources/balance.move
```diff
@@ -1,45 +1,59 @@
...
+/// bounded by `0xFFFF * 0xFFFF < 2^32`.
+const MAX_UPPER_BOUND: u16 = 0xFFFF;
 
 // === Structs ===
 
-/// A single confidential amount: an `EncryptedAmount` plus the count of u16-bounded values that
-/// have been folded into it.
+/// One account's holdings of token `T`: the encrypted amounts, all under `pk`, plus
+/// `public_balance`, the plaintext total of public deposits credited to the pool but not yet merged.
 public struct EncryptedBalance<phantom T> has store {
```
benr-ml: how about Balances? and then active, pending, public

move/sources/balance.move
```diff
@@ -1,45 +1,59 @@
...
+    pk: PublicKey,
+    active: BoundedEncryptedAmount,
+    pending: BoundedEncryptedAmount,
+    public_balance: u64,
```
benr-ml, line 52: why not use here PublicCoin?

move/sources/balance.move
```diff
@@ -48,235 +62,344 @@ public struct EncryptedCoin<phantom T> {
...
+/// Issue a claim for `value`. The caller must have just moved matching funds into `T`'s pool.
+public(package) fun new_public_coin<T>(value: u64): PublicCoin<T> {
+    PublicCoin { value }
 }
 
-/// The verified encrypted amount carried by `coin`.
-public(package) fun amount<T>(coin: &EncryptedCoin<T>): &InRangeVerifiedEncryptedAmount {
-    &coin.amount
+/// Consume a claim, returning the amount to pay out of `T`'s pool.
+public(package) fun redeem_public_coin<T>(coin: PublicCoin<T>): u64 {
```
benr-ml: this one is risky since u64 has drop

move/sources/balance.move
```diff
@@ -48,235 +62,344 @@ public struct EncryptedCoin<phantom T> {
...
 }
 
-// === EncryptedBalance ===
+// === Deposits ===
 
-/// An `EncryptedBalance` encrypting zero with `upper_bound = 1`.
-public(package) fun new<T>(): EncryptedBalance<T> {
-    EncryptedBalance { amount: encrypted_amount::zero(), upper_bound: 1 }
+/// Add a `PublicCoin` to `self`.
+public(package) fun deposit_public<T>(self: &mut EncryptedBalance<T>, coin: PublicCoin<T>): u64 {
```
benr-ml:
```suggestion
public(package) fun deposit_public_and_get_amount_for_accounting<T>(self: &mut EncryptedBalance<T>, coin: PublicCoin<T>): u64 {
```

move/sources/balance.move
```diff
@@ -48,235 +62,344 @@ public struct EncryptedCoin<phantom T> {
...
 }
 
-// === PublicCoin / EncryptedCoin ===
-
-/// A `PublicCoin` of zero value.
-public(package) fun zero<T>(): PublicCoin<T> {
-    PublicCoin<T> { value: 0 }
+/// A batched transfer out of an `EncryptedBalance<T>`, verified and ready to execute.
+/// By construction, the sum of `receiver_amounts` is the total leaving the sender.
+public struct VerifiedTransfer<phantom T> {
```
benr-ml: as in the previous pr, it's a wrapper that mostly adds another layer of abstraction - all its fields are verified by definition, what does it solve?

jonas-lj: it does decouple a bit (the construction of the total_sender amount in particular), but it'd be better to just do everything inside try_withdraw_batch. This touches the auditor a bit though. I'll rewrite.

move/sources/balance.move
jonas-lj: not relevant anymore

move/sources/balance.move
jonas-lj: True. I've refactored and PublicCoin doesn't exist anymore

move/sources/balance.move
jonas-lj: It required a rewrite, but it's gone now

move/sources/balance.move
jonas-lj: That lacks some accuracy, I think. Perhaps ConfidentialBalance?

move/sources/session.move
jonas-lj: They're gone

move/sources/balance.move
benr-ml: i agree that Balance is more generic and it's not optimal, but those are indeed balances :)
your call...

benr-ml (approved):

PR #41, opened 2026-08-18
[move][guardian] add guardian pkg and approval check

joyqvq:


apps/guardian/move/sources/guardian.move
```diff
@@ -0,0 +1,407 @@
...
+/// The message an enclave must sign after checking the operation represented by `binding`.
+public fun new_guardian_request(binding: &Binding): GuardianRequest {
+    GuardianRequest { version: REQUEST_VERSION, digest: binding.digest() }
+}
+
+// === Public Functions: approvals called by client ===
+
+/// Return `none` when authority checks are disabled; otherwise verify that a registered enclave
+/// approved exactly this transfer and return its `Approval<T>`.
+public fun new_transfer_approval<T>(
```
benr-ml: 1. we better not have a function per flow which wallets need to work with (as they are custom per issuer)
2. why do we need to know it's a transfer here? I think it's enough to show a signed digest to get Approval, and the actual operation is checked in the relevant contra function (currently you call transfer_binding twice per tx, is that needed?)

move/sources/contra.move
```diff
@@ -266,6 +274,29 @@ public fun authorize_as_object<T>(ct: &ConfidentialToken<T>, uid: &mut UID): Aut
     policy::as_object<T>(&ct.policy, uid)
 }
 
+/// Mint the configured authority's approval of the operation with `digest`. Returns `none` when
+/// authority checks are disabled; otherwise aborts unless the configured witness is `W` and its
+/// object ID matches `authority_id`. Called by `guardian::new_transfer_approval` and
+/// `guardian::new_unwrap_approval`.
+public fun mint_approval<T, W: drop>(
+    ct: &ConfidentialToken<T>,
+    authority_id: ID,
```
benr-ml: 1. hmm, I think anyone can call this function with any ID
2. why do we need both the id and the witness?

joyqvq: 1- only guardian module can bc of the guardian witness
2- its needed bc one can create many guardian object, but just one is appointed

the guardian uid is stored with authoritycap and also approval now

apps/guardian/move/sources/guardian.move
joyqvq: good point, revised

move/sources/contra.move
benr-ml: >2- its needed bc one can create many guardian object, but just one is appointed
can you elaborate? when/why would there be more than 1?

joyqvq: i don’t mean that an issuer should normally create multiple Guardians. 

move does not enforce Guardian<T> as a singleton, so multiple instances can exist, e.g. during replacement. contra stores the authority_ref that defines the ID of the one guardian currently enabled for the token, and approval is bounded to that id

move/sources/contra.move
```diff
@@ -266,6 +274,20 @@ public fun authorize_as_object<T>(ct: &ConfidentialToken<T>, uid: &mut UID): Aut
...
 
+/// Mint the enabled authority's approval of the operation identified by `digest`. Guardian calls
+/// this from `guardian::new_approval` using its private `AuthorityCap<T>`, verifies the enclave
+/// signature, and only then returns the approval. Returns `none` when authority checks are disabled.
+public fun mint_approval<T>(
+    ct: &ConfidentialToken<T>,
+    cap: &AuthorityCap<T>,
+    digest: &vector<u8>,
+): Option<Approval<T>> {
+    if (ct.inner().authority_ref.is_none()) option::none()
```
benr-ml: should it be called when authority_ref.is_none()? or can we just fail in that case

move/sources/contra.move
```diff
@@ -916,6 +978,34 @@ public fun set_policy<T, W>(
...
+    _management_cap: &ManagementCap<T>,
+    authority_cap: &AuthorityCap<T>,
+) {
+    ct.inner_mut().authority_ref = option::some(authority::new_ref(authority_cap));
+    events::emit_authority_set<T>(authority::authority_id(authority_cap));
+}
+
+/// Unset the token's `AuthorityRef` to disable external authority checks so protected operations
+/// no longer require an `Approval<T>`.
+public fun unset_authority_ref<T>(ct: &mut ConfidentialToken<T>, _cap: &ManagementCap<T>) {
```
benr-ml: let's change set_authority_ref to receive option<AuthorityCap> and just set it as is

move/sources/contra.move
```diff
@@ -538,13 +562,30 @@ public fun batched_transfer<T>(
...
 ): TransferBatch<T> {
     ct.assert_token_active(deny_list);
     let sender_addr = sender.owner();
     assert!(auth.is_authenticated(sender_addr), EAuthorizationError);
     assert!(!is_sender_denied<T>(deny_list, sender_addr), ETransferDenied);
     assert!(!receiver_amounts.is_empty(), EEmptyTransferBatch);
     assert!(receiver_amounts.length() <= MAX_BATCH_RECIPIENTS, EBatchTooLarge);
+    let token = &sender[TokenAccountKey<T>()];
+    if (ct.inner().authority_ref.is_none()) {
+        authority::discard_approval(approval);
```
benr-ml: assert(approval.is_none(), ..), no?

move/sources/authority.move
```diff
@@ -0,0 +1,149 @@
...
+
+/// Explicitly consume and discard an optional approval when authority checks are disabled.
+public(package) fun discard_approval<T>(approval: Option<Approval<T>>) {
+    approval.do!(|approval| {
+        let Approval { authority_id: _, digest: _ } = approval;
+    });
+}
+
+/// Consume `approval` against the enabled authority and operation `binding`.
+public(package) fun consume_approval<T>(
```
benr-ml: public(package) fun consume<T>(
    approval: Approval<T>,
    authority_ref: &AuthorityRef,
    binding: Binding)

and then call it below using approval.consume(...)

apps/guardian/move/sources/guardian.move
```diff
@@ -0,0 +1,372 @@
...
+    version: u16,
+    min_version: u16,
+    pcrs: Pcrs,
+}
+public struct EnclaveRegisteredEvent<phantom T>(GuardianEnclaveKey) has copy, drop;
+public struct EnclaveRemovedEvent<phantom T>(GuardianEnclaveKey) has copy, drop;
+
+// === Public Functions: constructors ===
+
+public fun new_guardian_signature(key_index: u8, signature: vector<u8>): GuardianSignature {
```
benr-ml: why do we need this intermediate step/struct?

move/guardian/sources/guardian.move
```diff
@@ -0,0 +1,372 @@
...
+        authority_cap,
+        operator,
+        url: b"".to_string(), // set by operator later.
+        version: 0,
+        min_version: 0,
+        pcrs,
+        guardian_enclave_keys: vec_map::empty(),
+    };
+    guardian.emit_updated();
+    transfer::share_object(guardian);
```
benr-ml: better return it and let the caller call share_obj

apps/guardian/move/sources/guardian.move
```diff
@@ -0,0 +1,372 @@
...
+        pcrs,
+        guardian_enclave_keys: vec_map::empty(),
+    };
+    guardian.emit_updated();
+    transfer::share_object(guardian);
+}
+
+/// Enable or replace this Guardian as the token's authority. The issuer must enable it here because
+/// only the Guardian module can borrow the `AuthorityCap<T>` stored in the Guardian's private field.
+public fun set_authority<T>(
```
benr-ml: why do we need to replace the guardian? 
can't we just update its fields?

apps/guardian/move/sources/guardian.move
```diff
@@ -0,0 +1,372 @@
...
+    /// The URL that fronts the enclave fleet. The operator can update it.
+    url: String,
+    /// Incremented per PCR change.
+    version: u16,
+    /// Keys registered below this version are pruned.
+    min_version: u16,
+    /// The expected enclave image measurements.
+    pcrs: Pcrs,
+    /// The fleet's enclave keys, keyed by a slot index.
+    guardian_enclave_keys: VecMap<u8, GuardianEnclaveKey>,
```
benr-ml: something we should document somewhere - to not break applications like payment channels, i think the issuer would be encouraged to not remove immediately old enclaves but only after, say, 1 day, to not break already signed approvals.

move/sources/authority.move
```diff
@@ -0,0 +1,149 @@
...
+/// The arguments an `Approval` commits to. Every variant includes the sender's key, the exact
+/// encrypted limbs of the active balance, and the operation arguments, so an approval is void once
+/// any of them changes.
+public enum Binding has drop {
+    Transfer {
+        sender_pk: PublicKey,
+        receiver_pks: vector<PublicKey>,
+        old_encrypted_balance: EncryptedAmount,
+        new_encrypted_balance: EncryptedAmount,
+        encrypted_amounts: vector<EncryptedAmount>,
```
benr-ml:
```suggestion
        receiver_encrypted_amounts: vector<EncryptedAmount>,
```

move/guardian/sources/guardian.move
```diff
@@ -0,0 +1,372 @@
...
+    signing_pk: Ed25519PublicKey,
+    /// Registered encryption public key for client to seal request to enclave.
+    enc_pk: X25519PublicKey,
+    /// Guardian version at registration; pruned once `min_version` is higher.
+    version: u16,
+}
+
+/// The message an enclave signs: protocol version and the digest of the checked operation.
+public struct GuardianRequest has copy, drop {
+    version: u16,
```
benr-ml: what do we expect to do with version?

move/sources/contra.move
joyqvq: the approval is a hot potato so needs to be consumed. so if someone submitted an approval but its disabled it can still pass as a noop. treating it as an error may be too harsh on timing difference

move/guardian/sources/guardian.move
joyqvq: when the payload from guardian changes.

apps/guardian/move/sources/guardian.move
joyqvq: this is just an initial setup (set or unset will enable/disable the guardian). you do not need to replace guardian and just need to update it

discussed offline: change to a singleton construction

move/sources/contra.move
joyqvq: returns none here is useful not to fail the entire ptb, for example guardian is just disabled and the txn with an approval is submitted immediately, the txn should pass (with a slight waste of gas) instead of failing to client and retry. this helps with these transient issues imo. its client's job to refresh the onchain state once in a while to decide whether new_approval is needed in the ptb, since we expect the guardian enable/disable should be an infrequent event.

move/sources/contra.move
joyqvq: discussed offline guardian will be a single created claiming the registry id so one token<T> can only create one guardian<T>

move/sources/contra.move
joyqvq: i think the separate unset_authority name is clearer because this entry point cannot enable an authority. enabling must go through guardian::set_authority, and requires `AuthorityCap`. the only issuer-facing function can only disable the configured authority.

```
// this is called by guardian::set_authority
public fun set_authority<T>(
    ct: &mut ConfidentialToken<T>,
    _management_cap: &ManagementCap<T>,
    authority_cap: &AuthorityCap<T>, // the api here is different
)

public fun unset_authority<T>(
    ct: &mut ConfidentialToken<T>,
    _management_cap: &ManagementCap<T>,
)
```

move/sources/authority.move
joyqvq: revised

move/guardian/sources/guardian.move
joyqvq: revised and added comments

apps/guardian/move/sources/guardian.move
joyqvq: added to comments for removes

apps/guardian/move/sources/guardian.move
joyqvq: good point, removed and inlined to to the transfer/unwrap api

joyqvq:
@benr-ml see latest commit addresses most of your comments and includes some gas optimizations, event changes, and API renames. PTAL 🙏 

for Guardian, I think we can make the canonical guardian pkg part of the framework. once framework is deployed it creates one shared GuardianRegistry. each issuer would then instantiate the guardian config for its token type through that registry. the registry reserves a derived-object slot keyed by GuardianKey<T>, so there is one canonical Guardian<T> per token type.

```
guardian::new_guardian<T>(
    &mut registry,
    &management_cap,
    pcrs,
    operator,
)
```

Issuers can use this canonical Guardian with free-tier support out of the box. They can still deploy a custom Guardian or authority package, but it would require separate allowlisting and would be subject to the consensus-path activation delay.

the impl on this pr can stay as is for decoupling still since guardian is over T.

apps/guardian/move/sources/guardian.move
```diff
@@ -0,0 +1,403 @@
...
+}
+
+/// Registered enclave keys.
+public struct GuardianEnclaveKey has copy, drop, store {
+    /// Index in `guardian_enclave_keys`.
+    index: u8,
```
benr-ml: is it needed?

move/sources/authority.move
```diff
@@ -0,0 +1,158 @@
...
+// === Errors ===
+
+const EApprovalMismatch: u64 = 2;
+const EWrongAuthority: u64 = 3;
+
+// === Types ===
+
+/// The external authority currently enabled for the token.
+public struct Authority has drop, store {
+    authority_id: ID,
```
benr-ml, line 41: do we still need this?

apps/guardian/move/sources/guardian.move
```diff
@@ -0,0 +1,403 @@
...
+public struct Pcrs(vector<u8>, vector<u8>, vector<u8>) has copy, drop, store;
+
+/// 32-byte ed25519 public key.
+public struct Ed25519PublicKey(vector<u8>) has copy, drop, store;
+
+/// 32-byte X25519 public key for sealing requests.
+public struct X25519PublicKey(vector<u8>) has copy, drop, store;
+
+/// Registry that reserves one derived Guardian object slot per token type.
+public struct GuardianRegistry has key { id: UID }
```
benr-ml: this requires a special init in the validators - if the guardian pkg is part of the contra package then it'd be cleaner to create Guardian as a derived address of its confidentialToken, no?

move/sources/authority.move
joyqvq: currently we keep the authority generic, so we still assume here that there can be more than one authority per token thats not guardian. so this id is still necessary to identify that. useful to keep it flexible so we want to enable some other authority instance in the future.

apps/guardian/move/sources/guardian.move
joyqvq: good point, looks like validator wont execute this init. revised to use derived address

apps/guardian/move/sources/guardian.move
joyqvq: yes we can remove this. 

while looking at this, we can also store this as vector<Option<GuardianEnclaveKey>> to keep the lookup cheap on hot path at o(1) whereas vecmap is o(n). the size should be small tho if we cap it at 16 (normally only 8)

move/sources/authority.move
benr-ml: > useful to keep it flexible so we want to enable some other authority instance in the future.

why isn't the AuthorityCap enough for that?

joyqvq: maybe I’m missing something, but when a client transfers or unwraps, it first gets an Approval<T> from an authority that contains the minting authority’s ID. When Contra consumes it, it checks that approval.authority_id == token.authority.authority_id. 

the  AuthorityCap<T> is private to the authority impl, why is this enough?

joyqvq: scenario: 
1. issuer enables the canonical Guardian<T>, which holds one AuthorityCap<T>.
2. later the issuer enables a different authority (or a custom guardian pkg deploy) with another AuthorityCap<T>.
3. if confidentialtoken<T> doesnt store the authority id, it cannot distinguish the old vs new authority. 

currently we only restrict singleton per token for the canonical guardian pkg. we do not restrict other authority impls or custom guardian impls. unless we want that, we should keep the authority id in struct authority

move/guardian/sources/guardian.move
```diff
@@ -0,0 +1,392 @@
...
+/// once per token. The Guardian privately stores an `AuthorityCap<T>` bound to its derived object
+/// ID. To create and enable it atomically, call `enable_authority` before `share`; otherwise share
+/// it and enable it later.
+public fun new_guardian<T>(
+    ct: &mut ConfidentialToken<T>,
+    cap: &ManagementCap<T>,
+    pcr0: vector<u8>,
+    pcr1: vector<u8>,
+    pcr2: vector<u8>,
+    operator: address,
```
joyqvq: destroy old authority cap. 
token {
  enabled: bool
}

apps/guardian/move/sources/guardian.move
```diff
@@ -0,0 +1,392 @@
...
+    /// The URL that fronts the enclave fleet. The operator can update it.
+    url: String,
+    /// Incremented per PCR change.
+    version: u16,
+    /// Keys registered below this version are pruned.
+    min_version: u16,
+    /// The expected enclave image measurements.
+    pcrs: Pcrs,
+    /// The fleet's enclave keys, keyed by a slot index.
+    guardian_enclave_keys: VecMap<u8, GuardianEnclaveKey>,
```
joyqvq: vector<option<GuardianEnclaveKey>> capped at 64.

joyqvq:
cc @benr-ml i updated (squashed) all the changes to make it easier to review after design iterations. 

PTAL for the following changes: 
1. canonical guardian pkg. created with deterministically derived id exactly once so there is one canonical guardian per token. one guardian is created with one privately stored authority cap. 
2. contra holds option<ID>. enable sets this to the guardian id, and disable sets it to none. needs both mgmt cap and authority cap (from guardian)
3. mint_approval needs authority cap only and approval is consumed as a hot potato for transfer/unwrap (rekey and bal update not in this pr)
4. contra mgmt cap can be used to replace a canonical authority to some other instance (e.g. custom guardian) but no longer enforcing singleton.

move/guardian/sources/guardian.move
```diff
@@ -0,0 +1,406 @@
...
+/// `guardian::disable`. If enabled, the client must present an enclave-signed operation digest to
+/// `guardian::new_approval` to mint an approval, then pass the returned `Approval<T>` to the
+/// protected operation in the same PTB.
+///
+/// The issuer can update PCRs, minimum version, and the operator. Only the operator can register
+/// attested enclave keys; both the issuer and operator can remove them. The operator also sets the
+/// service URL. During planned key rotation, old and new enclave keys should overlap for at least as
+/// long as pending signatures may remain valid before the keys are removed. Raising `min_version`
+/// with `update` immediately prunes every older-version key.
+module guardian::guardian;
```
benr-ml: shouldn't this be part of the contra package now?

move/guardian/sources/guardian.move
```diff
@@ -0,0 +1,406 @@
...
+    /// Capability bound to this Guardian's UID using the issuer's management capability.
+    /// Privately stored in the guardian object.
+    authority_cap: AuthorityCap<T>,
+    /// The designated operator. The issuer can replace it using `ManagementCap<T>`.
+    operator: address,
+    /// The URL that fronts the enclave fleet. Set and updated by the operator.
+    url: String,
+    /// Incremented per PCR change.
+    version: u16,
+    /// Keys registered below this version are pruned.
```
benr-ml:
```suggestion
    /// Enclaves registered below this version are pruned on PCRs update.
```

move/guardian/sources/guardian.move
benr-ml: if the payload is changed, wouldn't it change the entire struct?
can you give an example of how a new request struct will be used here?

should we instead use an enum with V1 { digest }?

move/guardian/sources/guardian.move
```diff
@@ -0,0 +1,406 @@
...
+public fun new_approval<T>(
+    self: &Guardian<T>,
+    ct: &ConfidentialToken<T>,
+    digest: vector<u8>,
+    key_index: u8,
+    signature: vector<u8>,
+): Option<Approval<T>> {
+    ct.mint_approval(&self.authority_cap, &digest).map!(|approval| {
+        let key_index = key_index as u64;
+        assert!(key_index < MAX_GUARDIAN_ENCLAVE_KEYS, EEnclaveKeyNotRegistered);
```
benr-ml: subsumed by next line, can remove

move/guardian/sources/guardian.move
```diff
@@ -0,0 +1,406 @@
...
+}
+
+// === Public Functions: called by issuer ===
+
+/// Called by the issuer to create the confidential token's canonical Guardian with the expected
+/// PCRs and an operator. The `GuardianKey` derived-object slot under `ct` is claimed to ensure the
+/// Guardian is created once per confidential token by this package. The Guardian privately stores
+/// an `AuthorityCap<T>` bound to its derived ID. To create and enable it atomically, call `enable`
+/// before `share`; otherwise share it and enable it later.
+public fun new_guardian<T>(
```
benr-ml: let's change to (private) entry all functions that are not needed as public

move/sources/authority.move
```diff
@@ -0,0 +1,170 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+/// A confidential token can enable one external authority at a time. While an authority is enabled,
+/// each protected operation must carry an `Approval<T>` minted using
+/// its privately stored `AuthorityCap<T>` over the operation digest. This mechanism is independent
+/// of the token's `Policy`.
+module contra::authority;
```
benr-ml, line 25: nit - given that we name this `authority`, perhaps we should rename guardian to NitroAuthority?

move/sources/authority.move
```diff
@@ -0,0 +1,170 @@
...
+    Binding::Unwrap {
+        sender_pk,
+        old_encrypted_balance,
+        new_encrypted_balance: *new_balance,
+        amount,
+    }
+}
+
+/// Return the canonical Blake2b-256 digest of the BCS-encoded operation binding.
+public fun digest(binding: &Binding): vector<u8> {
```
benr-ml:
```suggestion
fun digest(binding: &Binding): vector<u8> {
```

move/guardian/sources/guardian.move
benr-ml: i feel that i've already suggested that somewhere, so sorry if we already discussed it. if it's part of contra then a possible further simplification:
- a guardian can only be constructed via new_guardian using the management cap -> so it should also be able to mint Approval directly, without an AuthorityCap; this can be done by the guardian function having access to public(package) function from contra
- then, we don't need to store authority id anymore, no? instead of authority_id: Option<ID> we could have AuthorityType enum { Guardian, Custom{id} }, where for Custom the AuthorityCap is required

move/guardian/sources/guardian.move
joyqvq: this function returns the guardian object (that can be composed later to share or enable) so it needs to be public

its restricted that `a composable non-drop value must come from a public fun.` (from the move doc)

move/guardian/sources/guardian.move
joyqvq: addressed!

joyqvq:
added more changes as discussed today cc @benr-ml 

- canonical guardian is now nitro_authority.move in contra
- contra holds Option<AuthorityKind>, supports canonical Nitro and Custom { id } authorities. none means disabled.
- enable/disable flow called with management cap in contra only
- new_approval in nitro_authority just mints with the canonical nitro_authority object
- new_approval for a custom authority impl can call contra for mint_custom_authority_approval with authority cap. and it checks if contra's authoritykind::custom.id matches the authority_cap.authority_id
- mint in contra has public(package) for nitro_authority to call, and public mint for custom authority to call with authority cap. 
- same as before: nitro authority can only be created once for the token with derived id. custom authority can be many.

move/guardian/sources/guardian.move
joyqvq: good point, revised

apps/guardian/move/sources/guardian.move
joyqvq: changed to vector of options here with max = 64

move/sources/contra.move
```diff
@@ -161,6 +168,8 @@ public struct ConfidentialTokenV1 has drop, store {
     freeze_admins: VecSet<address>,
     policy: Option<Policy>,
     auditors: Auditors,
+    /// The enabled authority; `none` means authority checks are disabled.
+    authority: Option<AuthorityKind>,
```
benr-ml: add None to AuthorityKind instead of having Option here

move/sources/authority.move
```diff
@@ -0,0 +1,170 @@
...
+    /// Contra's canonical Nitro enclave authority.
+    Nitro,
+    /// A custom authority implemented by another package.
+    Custom { id: ID },
+}
+
+/// A one-use hot-potato approval bound to the enabled authority and one operation digest. It must
+/// be consumed by a protected operation in the same PTB.
+public struct Approval<phantom T> {
+    authority: AuthorityKind,
```
benr-ml: why do we need authority?

move/sources/contra.move
```diff
@@ -941,6 +1065,30 @@ public fun update_auditors<T>(
...
+    })
+}
+
+/// Borrow the confidential token's UID to claim its canonical derived Nitro authority.
+/// `nitro_authority::new` calls this package-only function using the issuer's `ManagementCap<T>`,
+/// which restricts creation to the issuer.
+public(package) fun authority_parent<T>(
+    ct: &mut ConfidentialToken<T>,
+    _management_cap: &ManagementCap<T>,
+): &mut UID {
```
benr-ml, line 1063: oh, didn't know this is possible (to return &mut UID)

move/sources/nitro_authority.move
```diff
@@ -0,0 +1,386 @@
...
+/// The versioned message an enclave signs. Add a new variant when the signed payload changes.
+public enum NitroAuthorityRequest has copy, drop {
+    V1 { digest: vector<u8> },
+}
+
+// === Events ===
+
+/// The canonical `NitroAuthority<T>` core configuration, emitted on creation and after `update` or
+/// `set_url`.
+public struct NitroAuthorityUpdatedEvent<phantom T> has copy, drop {
```
benr-ml: let's move all events to events.move now that it's 1 package

move/sources/nitro_authority.move
```diff
@@ -0,0 +1,386 @@
...
+        approval
+    })
+}
+
+// === Entry Functions: called by issuer ===
+
+/// Share a newly created `NitroAuthority<T>`. The issuer calls this entry function after `new`,
+/// optionally after enabling Nitro in the same PTB; ownership of the unshared object authorizes the
+/// call.
+entry fun share<T>(nitro_authority: NitroAuthority<T>) {
```
benr-ml: must be public

move/sources/nitro_authority.move
```diff
@@ -0,0 +1,386 @@
...
+
+// === Entry Functions: called by operator ===
+
+/// Register an enclave whose attestation document matches the Nitro authority's PCRs. The operator
+/// calls this entry function, and `TxContext.sender` must equal the configured operator; no
+/// capability is required. The function parses `signing_pk || enc_pk` from `user_data` and stores
+/// the key pair in the lowest free slot.
+entry fun register_enclave<T>(
+    self: &mut NitroAuthority<T>,
+    document: NitroAttestationDocument,
```
benr-ml: i don't think it can receive NitroAttestationDocument created by an earlier ptb command as an entry function

move/sources/authority.move
joyqvq: good point we can remove since its a hot potato, just checking it when and consuming is sufficient

move/sources/contra.move
joyqvq: good point, this simplify some code..

move/sources/authority.move
joyqvq: yes since its a hot potato we can remove this

move/sources/contra.move
```diff
@@ -548,16 +559,35 @@ public fun batched_transfer<T>(
...
     ct.assert_token_active(deny_list);
     let sender_addr = sender.owner();
     assert!(auth.is_authenticated(sender_addr), EAuthorizationError);
     assert!(!is_sender_denied<T>(deny_list, sender_addr), ETransferDenied);
     assert!(!receiver_amounts.is_empty(), EEmptyTransferBatch);
     assert!(receiver_amounts.length() <= MAX_BATCH_RECIPIENTS, EBatchTooLarge);
+    let token = &sender[TokenAccountKey<T>()];
+    assert!(!token.is_frozen, ETransferDenied);
+    let inner = ct.inner();
+    if (authority::is_none(&inner.authority)) {
```
jonas-lj: In order to not bloat contra.move, could we move this logic to the authority module and just call `inner.authority.verify(approval)` and then handle the cases (Some/None x authority kinds) there?

move/sources/contra.move
```diff
@@ -814,6 +851,22 @@ fun try_unwrap_internal<T>(
     assert!(!is_sender_denied<T>(deny_list, owner), ETransferDenied);
     let account = &mut account[TokenAccountKey<T>()];
     assert!(!account.is_frozen, ETransferDenied);
+    let inner = ct.inner();
+    if (authority::is_none(&inner.authority)) {
```
jonas-lj: Same as above

move/sources/contra.move
joyqvq: I added this branch intentionally so the operation binding can be lazy. 

if we use authority.verify(approval, binding) helper would require constructing and copying the transfer binding even when AuthorityKind::None

joyqvq: we can perhaps add a macro for this though

move/README.md
```diff
@@ -1,6 +1,27 @@
-# Move package
+# Move packages
 
-On-chain Move modules for the project. See the [top-level README](../README.md) for what they do.
+This directory contains the Contra Move package. See the [top-level README](../README.md) for what it does.
+
+## Authority flow
```
benr-ml: let's keep this readme super short and move this documentation to authority.move

README.md
```diff
@@ -74,6 +74,7 @@ flowchart LR
 
 - **[`move/`](move/)** -- on-chain Move contracts, including:
   - **[`contra.move`](move/sources/contra.move)** -- main entry point with the high-level interfaces for both issuers and users; see the module-level doc comment at the top of the file for the full list of flows.
+  - **[`nitro_authority.move`](move/sources/nitro_authority.move)** -- `nitro_authority` is Contra's canonical authority for protected operations, backed by AWS Nitro enclaves. Each issuer can create one derived `NitroAuthority<T>` for its `ConfidentialToken<T>`, enable or disable it, and configure its PCRs, operator, and enclave keys.
```
benr-ml: let's remove this as contra explains the high leve

README.md
```diff
@@ -175,6 +176,11 @@ In addition to the steps above, the issuer can gate selected user flows behind t
 
 See the [closed-loop app](apps/closed-loop/) for an example that gates `register` behind a whitelist.
 
+#### Approval authority setup
+
+Contra can require authority approval as a second factor in addition to the zero-knowledge proofs required by protected operations. `ConfidentialToken<T>` stores an `AuthorityKind`: `None` disables authority checks, `Nitro` selects the canonical `NitroAuthority<T>`, and `Custom { id }` selects another authority implementation. The `ManagementCap<T>` holder calls `contra::enable_authority` with `AuthorityKind::Nitro` or `AuthorityKind::Custom { id }` to enable or replace the authority, and calls `contra::disable_authority` to set it to `AuthorityKind::None`.
```
benr-ml: let's write it without too much technical parts, e.g., - The issuer may define an authority for its token, either based on AWS Nitro or customized. 

also below, let's just explain the flow in high level

move/sources/contra.move
joyqvq: revised to use macro!

move/sources/contra.move
```diff
@@ -926,6 +968,74 @@ public fun set_policy<T, W>(
...
+// - For both the canonical Nitro authority and custom authorities, protected operations require an
+//   approval while the authority is enabled and consume it against the operation binding.
+
+/// Enable `new_authority` for protected operations. The issuer calls this public function directly
+/// using its `ManagementCap<T>`. Enabling a new authority replaces the current authority; enabling
+/// the active authority is a no-op.
+public fun enable_authority<T>(
+    ct: &mut ConfidentialToken<T>,
+    _management_cap: &ManagementCap<T>,
+    new_authority: AuthorityKind,
```
jonas-lj, line 994: I think we need a constructor for this. All constructors are currently test-only.

move/sources/authority.move
```diff
@@ -0,0 +1,213 @@
...
+}
+
+/// A one-use hot-potato approval bound to one operation digest. It must be consumed by a protected
+/// operation in the same PTB.
+public struct Approval<phantom T> {
+    digest: vector<u8>,
+}
+
+/// The arguments an `Approval` commits to.
+/// TODO: add rekey and balance update.
```
joyqvq, line 64: add comment that why the token and the id are agnostic

move/sources/nitro_authority.move
```diff
@@ -0,0 +1,370 @@
...
+/// `Approval<T>`. Contra reconstructs the same digest when consuming the approval.
+public fun new_approval<T>(
+    self: &NitroAuthority<T>,
+    ct: &ConfidentialToken<T>,
+    digest: vector<u8>,
+    key_index: u8,
+    signature: vector<u8>,
+): Option<Approval<T>> {
+    ct.mint_nitro_authority_approval(&digest).map!(|approval| {
+        let key_index = key_index as u64;
```
jonas-lj, line 139: nit: calling with key_index = 63 gives EEnclaveKeyNotRegistered (assuming that slot is empty) but key_index = 64 gives a vector error.

move/sources/contra.move
```diff
@@ -926,6 +968,74 @@ public fun set_policy<T, W>(
...
+    if (*authority == new_authority) return;
+    if (!authority::is_none(authority)) events::emit_authority_disabled<T>(*authority);
+    *authority = new_authority;
+    if (!authority::is_none(authority)) events::emit_authority_enabled<T>(new_authority);
+}
+
+/// Disable authority checks by setting the configured authority to `AuthorityKind::None`. The
+/// issuer calls this public function directly using its `ManagementCap<T>`. Calling it while
+/// `AuthorityKind::None` is already configured is a no-op.
+public fun disable_authority<T>(ct: &mut ConfidentialToken<T>, _management_cap: &ManagementCap<T>) {
```
jonas-lj: Isn't this just enable_authority with `new_authority` equal to none?

jonas-lj: not regarding the event type

move/sources/nitro_authority.move
joyqvq: are you suggesting to add a bound check? leaving as is without it is cheaper on the fast path

move/sources/contra.move
joyqvq: looks like AuthorityKind can be bcs-serialized and passed directly as a pure ptb arg so no need for constructor (to save gas too) https://sdk.mystenlabs.com/bcs

move/sources/contra.move
joyqvq: sure we can combine into set authority and account for it in event type

move/sources/nitro_authority.move
jonas-lj: Yes, I'd add an assert!(key_index < MAX_NITRO_AUTHORITY_ENCLAVE_KEYS, EEnclaveKeyNotRegistered) to get a consistent error message. The cost is one u64 comparison which is negligible. It's only about the error a client sees, so fine to skip.

PR #43, opened 2026-08-20
[Move] Rename the verified-amount types

jonas-lj:
Renames the verified-amount type ladder so the names state the guarantee rather than the construction: `BoundedEncryptedAmount` → `AccumulatedAmount`, `VerifiedEncryptedAmount` → `VerifiedAmount`, `InRangeVerifiedEncryptedAmount` → `RangeVerifiedAmount`.


benr-ml (approved):
pls update kaisho binary

PR #45, opened 2026-08-20
[Move] Unify proof parameter names on the `pok` postfix and fix misleading error codes

jonas-lj:
Two naming fixes on the public Move boundary.

**Proof parameters** — the same argument was `new_balance_pok` in `update_active_balance` (and throughout `balance.move`) but `new_balance_consistency_proof` in `unwrap` / `try_unwrap`. Unified on the `pok` postfix already used elsewhere. `receiver_encs_pok` / `sender_encs_pok` were already consistent and are untouched.

**Error codes** — code 0 was raised both when an `Account` already existed and when a `TokenAccount<T>` did; `EAmountsEqualityProofFailed` was the abort for a failed re-key proof. Split out `ETokenAccountAlreadyRegistered` (code 4), renamed the two misleading constants (values unchanged), documented why 10 is free.


benr-ml (approved):
pls update kaisho binary

PR #50, opened 2026-08-23
[Move, TS-SDK] clean ups

benr-ml:
Mostly shorter docs + small code simplifications. also,
- wrap and unwrap reject zero amounts
- new_auditor_package enforces every handle pair is exactly [lo, hi]
- Auditors::new sets previous_pks=[]
- many functions narrowed to public(package)
- all TS transaction builders are now async (consistency)
- updated the payment channel example


move/sources/policy.move
```diff
@@ -44,11 +44,14 @@ public struct Auth<phantom T> has drop {
 /// operation index exceeds `MAX_OPERATION_INDEX` or if `permissioned_operations` contains
 /// duplicates.
 fun new<W>(permissioned_operations: vector<u8>): Policy {
-    assert!(permissioned_operations.all!(|o| *o <= MAX_OPERATION_INDEX), EInvalidOperation);
-    assert!(is_unique(&permissioned_operations), EDuplicateOperation);
+    let bitmap = permissioned_operations.fold!(0u32, |acc, o| {
+        assert!(o <= MAX_OPERATION_INDEX, EInvalidOperation);
+        assert!(acc & (1 << o) == 0, EDuplicateOperation);
```
jonas-lj: nit: 1 << o is computed twice

move/sources/balance.move
```diff
@@ -287,6 +286,26 @@ fun verify_transfer_amounts<T>(
...
+    session_id: SessionId,
+): bool {
+    let new_balance = self.verify_amount(
+        new_balance,
+        new_balance_pok,
+        new_balance_range_proofs,
+        session_id,
+    );
+    let mut residual = self.balance_change(&new_balance);
+    residual.add_assign_u64(amount);
```
jonas-lj, line 306: Add a guard for amount > 0. Otherwise `try_update_active`, which calls this with 0, is more expensive.

move/sources/decode.move
```diff
@@ -1,8 +1,8 @@
 // Copyright (c) Mysten Labs, Inc.
 // SPDX-License-Identifier: Apache-2.0
 
-/// Simple deserialization functions that build the composite crypto types from their
-/// byte-encoded elements in a single Move call.
+/// Optional convenience deserializers that build the composite crypto types from their byte-encoded
+/// elements in a single Move call. Do not check the input validity.
```
jonas-lj, line 5: nit: The group and scalar encodings are checked, but everything else isn't.

jonas-lj (approved):

move/sources/policy.move
jonas-lj: The `1 << o` computation still has to be after line 48 though.

PR #51, opened 2026-08-24
[Move] Make ConfidentialToken's state a versioned enum

jonas-lj:


move/sources/contra.move
```diff
@@ -317,19 +330,19 @@ public fun share_account(account: Account) {
 /// which must be for the `PERMISSIONED_REGISTER` operation and for `account.owner`.
 public fun register<T>(account: &mut Account, auth: &Auth<T>, pk: PublicKey) {
     assert!(auth.is_allowed(PERMISSIONED_REGISTER), EAuthorizationError);
-    assert!(auth.is_authenticated(account.owner), EAuthorizationError);
+    assert!(auth.is_authenticated(account.owner()), EAuthorizationError);
```
benr-ml, line 341: an alternative pattern to having getters is:
let account = account.inner(); // or let ct = ct.inner();
... use the previous code ...
that option reduces the number of function calls and the number of match evals

benr-ml (approved):

PR #58, opened 2026-08-25
[guardian] add cli crate for enclave devops

joyqvq:


guardian/cli/src/main.rs
```diff
@@ -0,0 +1,193 @@
...
+#[derive(Subcommand)]
+enum Command {
+    /// Manage a Guardian as the confidential token issuer.
+    Issuer {
+        /// Issuer address used as the transaction sender.
+        #[arg(long, env = "SUI_SENDER")]
+        sender: Address,
+        /// Sign and execute with the issuer key in the configured wallet.
+        #[arg(long)]
+        execute: bool,
```
benr-ml, line 54: since all commands here are sensitive, should we just focus on returning unsigned tx (and remove wallet, etc)?

benr-ml (approved):

PR #66, opened 2026-09-03
[Docs] State that RPC-fetched encryption keys are unauthenticated

jonas-lj:
Document that the receiver and auditor keys the SDK encrypts transfer amounts under come from unauthenticated RPC reads, so the endpoint is part of the user's trust base.


benr-ml (commented):
i think it's obvious, no need to add those docs

